PR #109, opened 2014-04-11
Fix signed/unsigned warnings

roystgnr:
Do we have decent test coverage of SplitString()?  If not then don't bother to merge this yet; I was just getting sick of the compiler warnings, but that's not worth risking a regression for.


roystgnr:
Superceded by #284

PR #130, opened 2014-06-18
Verbose mesh construction

roystgnr:
This was useful during some mesh refinement debugging; it might be useful in the future too if you like.


roystgnr:
This is so out of date now it's not even funny.

PR #145, opened 2014-09-22
Allow compact specification of multi-bcid BCs

roystgnr:
This should let us specify a multi-bcid as e.g. "14:15:16" rather than
writing multiple separate ids with "14", "15", "16" and redundant
information.

This patch needs some code review and testing before merging.


pbauman:
I'm closing this since we refactored the boundary condition parsing and it naturally supports multiple bc ids.

PR #156, opened 2014-10-13
Velocity penalty orthogonality

roystgnr:
Not sure this is worth merging; putting it up for @nicholasmalaya to look at.

Nick, apparently the answer to the paradox of "if we can't impose any penalty force parallel to the velocity, then what happens if the penalty normal is parallel to the velocity" is "the nonlinear solver screams and dies".  The Jacobians here are satisfying numeric tests, I switched from ILU to MUMPS LU to get the linear solves to all converge, but the nonlinear solver dies after a few (barely residual reducing) steps anyway.  I'll let Bob know and we'll see if this is worth futzing with further.

### DIFF
--- a/src/physics/include/grins/velocity_penalty_base.h
+++ b/src/physics/include/grins/velocity_penalty_base.h
@@ -60,7 +60,12 @@ namespace GRINS
    
   protected:
 
+    // Make penalty force proportional to |u|(u*n) instead of just
+    // (u*n)
     bool _quadratic_scaling;
+
+    // Do not apply any force parallel to the velocity
+    bool _orthogonal_force;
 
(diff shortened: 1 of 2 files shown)


nicholasmalaya:
Just noticed this guy while poking around looking at older tickets-- but this is resolved via other issues (such as #232 ) and I believe we can kill this pull.

roystgnr:
This is actually orthogonal to #232.

(stares awkwardly at camera 3 while waiting for laugh track to die down)

This is an attempt to ensure that even a single VelocityPenalty does no work on the flow.  It doesn't seem to be a successful attempt in the worst-case scenario of flow that wants to go into the wall.

pbauman:
I'm going to go ahead and close this since it's been idle for awhile. If you guys decide this is worth it, please feel free to reopen or submit a new PR.

PR #160, opened 2014-11-03
VelocityPenalty2 to constrain velocities to 1D

roystgnr:
@nicholasmalaya, this should be straightforward to use: just add VelocityPenalty2 (and its AdjointStabilization) to the physics line and add a VelocityPenalty2 section for the vertical penalty_val.


roystgnr:
This branch is still useful when building against old libmesh, but it's way behind master.  Replacing this PR with an updated version in #196

PR #172, opened 2014-11-25
Velocity drag adjoint stabilization terms

roystgnr:
@nicholasmalaya, this should be ready for testing.

This shouldn't impact any of the physics that we're relying yet, so it's probably safe to merge even if the new features have bugs I missed.

In the above sentence, "relying yet" should read "relying on yet".  @nicholasmalaya, you should probably assume the new features have bugs I missed.

One bug I didn't miss because I want to hit it in one fell swoop: @vikramvgarg and I missed replacing the constant _mu with the Viscosity mu() in the VelocityPenaltyAdjoint classes; we'll need to do that before we can safely use them with more complicated transport models.


roystgnr:
Superceded by #198.

PR #173, opened 2014-11-25
Averaged fan adjoint

roystgnr:
Two reasons to take our time with this one:

1) It depends on #172 so we should merge that first.

2) The refactoring leads to a test failure with the *non* stabilized physics.  Upon closer examination I'm pretty sure that "regression" is a bug in the earlier version of the code (and the gold solution it generated), not a bug in the current code.  But at this point I wouldn't trust either version (nor the AveragedTurbine, which was based on the old code) until @nicholasmalaya is done setting up the analytic regression test we discussed.  The good news: although both codes look reasonable in the "eyeball norm" on the existing test case, if I understand the bug correctly the old code will scream obvious failure on our proposed test.


roystgnr:
Superceded by #199

PR #174, opened 2014-11-27
Averaged turbine adjoint

roystgnr:
One more for @nicholasmalaya to play with.  Branched from #173.

This is going to require some edits to the nonlocal_mass_residual method after we merge #159.


roystgnr:
Superceded by #200

PR #185, opened 2014-12-24
Elastic cable

cahaynes:
Pull request for elastic cable physics


pbauman:
@cahaynes I'm getting a failure on your test. I'm only matching to 4 digits and not 8. Are you sure you've committed the latest versions of the .xdr files for the suspended cable example?

cahaynes:
Corrected the .xdr file and adjusted the initial guess for the example to reduce the number of Newton iterations, takes roughly 300 now.

pbauman:
OK, test is passing for me now. Thanks!

PR #192, opened 2015-01-13
interior_rate only has the one version

roystgnr:
This compiles for me now.

Also passes "make check" on my laptop, but I'd appreciate a double-check on that; I'm still not sure what bitrot crippled my Sysnet stack while I was away.


pbauman:
Thanks for the fix. Sorry that crept in.

PR #193, opened 2015-01-14
Print perflog mid-simulation if requested

roystgnr:
This is a feature @nicholasmalaya may find useful - we output the accumulating PerfLog data every N timesteps, not just at the end of a simulation, so there's less of a wait time before being able to examine performance changes.

One oddity: due to the way the libMesh PerfLog works, we get the machine info with the first print, and not with any subsequent prints.  I think this is the best way to do things, but it looks weird and I can move the machine info back to the final PerfLog if that's preferred.


pbauman:
Neat. I think I'd prefer only having the machine info print once. It probably only looks weird cause we're used to seeing it that way all the time. Thanks!

PR #200, opened 2015-01-30
Averaged turbine adjoint new

roystgnr:
This supercedes #174

There's another regression test change for test_2d_pseudoprop, same bugfix as with the fan.  Getting a proper analytic test for these is high on my todo list now.


roystgnr:
Pulling this since the only thing questionable here is changes to already-questionable physics, and since we've got two other PRs that depend on it.

PR #201, opened 2015-01-30
Fix trailing whitespace in LICENSE

roystgnr:
Somehow that had gotten into some of the source files but not others.  They should all be clear now.

This PR should wait until after #200 - I branched from there to make sure I didn't miss the newly-added files.

We'll have another fix of this sort to go in after Spalart Allmaras


pbauman:
Thanks. After this is merged (and SA, etc.) we can update the year too.

roystgnr:
Hahaha - didn't even notice that.  Let me add that to this PR.

PR #205, opened 2015-02-04
Possible bugfix for a build failure on lonestar

roystgnr:
std::shared_ptr can be constructed from NULL, but the
boost::shared_ptr version that @nicholasmalaya is using apparently
cannot.  Both should have working default constructors, though.

### DIFF
--- a/src/physics/src/physics_factory.C
+++ b/src/physics/src/physics_factory.C
@@ -150,7 +150,7 @@ namespace GRINS
         (new Subclass<ParsedViscosity>(physics_to_add,input));
  
     visc_error(physics_to_add, viscosity);
-    return PhysicsPtr(NULL);
+    return PhysicsPtr();
   }
 
   template <template<typename> class Subclass>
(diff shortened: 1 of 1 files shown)


roystgnr:
In theory, no difference.  In practice, I'm not even sure why you're seeing an error from boost::shared_ptr (we specifically typedef std::tr1::shared_ptr in physics_factory.h), nor why boost isn't just converting NULL to Physics* and using the constructor-from-pointer.

pbauman:
Might be a good prompt to wrap up shared_ptr. I'll try and find time over the weekend and do that.

nicholasmalaya:
Patch builds on lonestar, all tests are passing. 

<pre>
make[5]: Leaving directory `/work/00000/npm7/src/roy_grins/build/test'
============================================================================
Testsuite summary for grins 0.5.0
============================================================================
# TOTAL: 56
# PASS:  34
# SKIP:  22
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
</pre>

nicholasmalaya:
(Thank you for the quick turnaround, guys)

PR #206, opened 2015-02-04
Add test_2d_fantrick

roystgnr:
This sets up blades with no drag, at a 45 degree angle, moving 1 m/s
horizontally, in a flow moving 1 m/s vertically.  In other words,
blades which should do nothing.  And they do do nothing!  That's
verification, baby!


roystgnr:
Rebased with less ignorant vocabulary use in commit message.  This should be ready to merge whenever.

pbauman:
Moar tests!

PR #208, opened 2015-02-05
Copy, don't reference, physics_name strings

roystgnr:
Otherwise it's too easy to break Physics classes (as the current
PhysicsFactory code does) by passing in temporaries to their
constructors.


pbauman:
This is one of those not-well-thought-out-good-intentioned things. Sorry for that and thanks to @roystgnr for tracking it down.

pbauman:
Err, "this" being my original code.

PR #209, opened 2015-02-06
Postprocessing fixes

pbauman:
Fixes a couple of postprocessing bugs in the ElasticCable and ElasticMembrane Physics. Also fixed bug in the postprocessing if there were subdomain-restricted Physics enabled.

Will manually merge from 0.5.0-release into master.


pbauman:
The Physics changes don't impact anybody outside of my group at the moment and I've double checked that output for the default all subdomains case works so I'm going to go ahead and merge this.

PR #212, opened 2015-02-16
Guard against get_elem() calls for SCALAR variables

pbauman:
Basically in conjunction with libMesh/libmesh#467. With that plus this PR, master `make check` runs cleanly.


roystgnr:
Wouldn't adding a has_elem() check to _general_residual be better than forking it into _nonlocal_residual?

pbauman:
Fair enough. I was 6-in-one/half-dozen-the-other about it, but you're probably right. Updated accordingly. `make check` passing.

roystgnr:
Looks good to me; thanks.

pbauman:
Cool, thanks. I'll merge this after the libMesh PR.

PR #214, opened 2015-02-16
Doxygen update

pbauman:
Fixes Doxygen input file to actually build documentation for all the classes. This required 1. updating the INPUT variable (just tell it look in src and it will find everything) and 2. actually including Antioch and Cantera. We could've accomplished 2 by cluing Doxygen into grins_config.h in the build directory where then it would only include Antioch, for example, if it was actually present in the build. Using the PREDEFINED variable, we can tell Doxygen to just define those preprocessing parameters so that their documentation is always built. I prefer this so that we just always have the documentation for everything built all the time.

Set to merge to 0.5.0-release, will manually merge to master after this PR is merged.

Will leave this up for comments tonight and then merge tomorrow if no one has anything to add.

### DIFF
--- a/doxygen/grins.dox.in
+++ b/doxygen/grins.dox.in
@@ -1263,7 +1263,8 @@ INCLUDE_FILE_PATTERNS  =
 # undefined via #undef or recursively expanded use the := operator 
 # instead of the = operator.
 
-PREDEFINED             = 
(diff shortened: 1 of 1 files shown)


pbauman:
@roystgnr mentioned we should make sure and document in the optional library headers that they are configure time optional even though the documentation will always be built. I agree and will do that in a separate PR.

PR #220, opened 2015-02-19
Symmetry Boundary Conditions for Solid Mechanics/Mixed Dimension Example

pbauman:
Added symmetry boundary conditions for solid mechanics problems. Also, added mixed dimension example (that uses symmetry boundary conditions) --- Mooney-Rivlin inflating sheet with 1D Hookean stiffener rods. Don't have the regression test yet as `libMesh::ExactSolution` needs updating to support mixed dimension meshes, but wanted to open this up to reference. Will merge once the regression test is in place.


pbauman:
@benkirk you might be interested in this picture.

mixed-dim-sheet

benkirk:
Intrigued, yes!  So are there mixed elements in that, or is this a symmetry BC breaking??

pbauman:
Mixed elements. Rubber sheet with two (much stiffer) elastic rods along the two edges that aren't fixed; symmetry enforced. (Imagine a square sheet with rods running along the x-axis and the y-axis.) The case where I simulate the full thing without symmetry enforced isn't quite doing the right thing (symmetry is breaking). I have a hunch where there problem is though - coming to that after the libMesh foo that's going on now.

PR #222, opened 2015-02-23
Updated Branch 'Inc_ns_sides'

nicholasmalaya:
Instead of trying to fast-forward the 'inc_ns_sides' branch to the latest GRINS, this circumvents (and replaces) that effort with a new merge request to master. Two files are update so that the branch will compile: inc_navier_stokes.C and inc_navier_stokes.h

The edits to make these function were minor, with a changelog approximately of:

* Adding template parameters (namely, ) to IncompressibleNavierStokes method element_time_derivative
* Added 'this' pointer to implicit parameter invocations, e.g. _dim --> this->_dim

The branch is compiling. However, DO NOT MERGE this pull request for master, because I am seeing regression test failures for nearly all inc_ns solutions at present. Worse than that, the solver is not converging for tests with the inc_ns solution.

I want this to be at least closer to master so I can get @roystgnr (and possible, @pbauman ) eyes on this or any recommendations on methods to stamp out bugs.

### DIFF
--- a/src/physics/include/grins/inc_navier_stokes.h
+++ b/src/physics/include/grins/inc_navier_stokes.h
@@ -64,6 +64,10 @@ namespace GRINS
 					  AssemblyContext& context,
 					  CachedValues& cache );
 
+    // Side Boundary Normal Flux Term (of form, \int d()/dn )
+    virtual void side_time_derivative( bool compute_jacobian,
+				       AssemblyContext& context,
+				       CachedValues& cache );
     // Constraint part(s)
     virtual void element_constraint( bool compute_jacobian,
 				     AssemblyContext& context,
--- a/src/physics/src/inc_navier_stokes.C
+++ b/src/physics/src/inc_navier_stokes.C
@@ -344,38 +344,39 @@ namespace GRINS
     return;
   }
 
-
-  void IncompressibleNavierStokes::side_time_derivative( bool compute_jacobian,
+  template<class Mu>
+  void IncompressibleNavierStokes<Mu>::side_time_derivative( bool compute_jacobian,
 							 AssemblyContext& context,
 							 CachedValues& /*cache*/ )
   {
 #ifdef GRINS_USE_GRVY_TIMERS
     this->_timer->BeginTimer("IncompressibleNavierStokes::side_time_derivative");
 #endif
+
     // We integrated (-grad(p),v)_\Omega by parts.
     // This equals (p,div(v))_\Omega - (div(pv),1)_\Omega
     // Which latter term equals (p,v*n)_d\Omega
     // In cases where v*n is not constrained away on the boundary we
     // need to include this term, so we integrate it here.
     // The number of local degrees of freedom in each variable.
-    const unsigned int n_u_dofs = context.get_dof_indices(_flow_vars.u_var()).size();
-    const unsigned int n_p_dofs = context.get_dof_indices(_flow_vars.p_var()).size();
+    const unsigned int n_u_dofs = context.get_dof_indices(this->_flow_vars.u_var()).size();
+    const unsigned int n_p_dofs = context.get_dof_indices(this->_flow_vars.p_var()).size();
     // Check number of dofs is same for _flow_vars.u_var(), v_var and w_var.
-    libmesh_assert (n_u_dofs == context.get_dof_indices(_flow_vars.v_var()).size());
-    if (_dim == 3)
-      libmesh_assert (n_u_dofs == context.get_dof_indices(_flow_vars.w_var()).size());
+    libmesh_assert (n_u_dofs == context.get_dof_indices(this->_flow_vars.v_var()).size());
+    if (this->_dim == 3)
+      libmesh_assert (n_u_dofs == context.get_dof_indices(this->_flow_vars.w_var()).size());
     // We get some references to cell-specific data that
     // will be used to assemble the linear system.
     // Element Jacobian * quadrature weights for side integration.
     const std::vector<libMesh::Real> &JxW =
-      context.get_side_fe(_flow_vars.u_var())->get_JxW();
+      context.get_side_fe(this->_flow_vars.u_var())->get_JxW();
     // The velocity shape functions (in global coords.)
     // at side quadrature points.
     const std::vector<std::vector<libMesh::Real> >& u_phi =
-      context.get_side_fe(_flow_vars.u_var())->get_phi();
+      context.get_side_fe(this->_flow_vars.u_var())->get_phi();
     // The pressure shape functions at interior quadrature points.
     const std::vector<std::vector<libMesh::Real> >& p_phi =
-      context.get_side_fe(_flow_vars.p_var())->get_phi();
+      context.get_side_fe(this->_flow_vars.p_var())->get_phi();
     const std::vector<libMesh::Point>& u_qpoint =
       context.get_side_fe(this->_flow_vars.u_var())->get_xyz();
     // Normal vectors on the side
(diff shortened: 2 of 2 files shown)


nicholasmalaya:
Commit ff5d71975eadc09f48b12564c40e564d4df0aea0 nicely summarizes the change between this pull and 'inc_ns_sides'

pbauman:
2 quick comments:

1. Doesn't matter now, but you could've used `git cherry-pick` to cherry-pick off @roystgnr's commits from the stale branch. (Just a future FYI.)

Much more importantly,

2. We need to think about how to handle this term. Because it shouldn't be included on boundaries where the velocities are constrained nor should it be included on outflow boundaries, when the outflow is the "standard" outflow condition because this term is part of that and if you integrate it back in, it's changing the boundary condition.

Do y'all have a case like that? I'm pretty sure none of the examples fall into this category.

nicholasmalaya:
I just chatted with @pbauman via phone, and wanted to capture some of the thoughts here. 

A principle point of confusion (which I have done thing to salve) is regarding the motivation for even including these terms. As I (now) understand it, pressure only appears in the normal part of the boundary terms, however, a natural boundary condition implicitly sets:

$ (-p + \n \mathbf{n} \cdot \nabla \mathbf{u} \cdot \mathbf{n}) = 0 $ on the boundary. Which would imply we do not need the additional terms associated with a call to side_time_derivative. However, I thought @roystgnr mentioned that was not being done for the case where we have inflow, as opposed to outflow, and that this terms was somehow necessary as a result. 

As an interesting aside, some tests that should not be failing, namely those with fully specified dirichlet boundary conditions, do appear to be failing, which might point toward a bug. In particular:

<pre>
PASS: test_thermally_driven_2d_flow.sh
FAIL: test_thermally_driven_3d_flow.sh
</pre>
Which is additionally confusing, given that the 2nd test is passing, while the 3d is failing. 

All the tests failing on my current build are:
<pre>
FAIL: test_ns_poiseuille_flow.sh
FAIL: test_axi_ns_poiseuille_flow.sh

...

FAIL: test_thermally_driven_3d_flow.sh
FAIL: test_2d_pseudofan.sh
FAIL: test_2d_pseudoprop.sh
FAIL: test_2d_fantrick.sh
FAIL: test_2d_proptrick.sh
FAIL: test_dirichlet_fem.sh

...

FAIL: backward_facing_step_regression.sh
FAIL: locally_refine_regression.sh
FAIL: penalty_poiseuille.sh
FAIL: penalty_poiseuille_stab.sh
FAIL: redistribute_regression.sh
FAIL: coupled_stokes_ns.sh
</pre>
The ellipses here specify there is a gap of tests that are passing, and are therefore cut, not that all the intervening tests are also failing.

roystgnr:
Any failures for tests with fully specified Dirichlet sides would suggest a bug, yes.  These terms add to the residual indices corresponding to boundary velocity degrees of freedom only, and those lines in the system ought to be wiped out by the application of Dirichlet constraints.

roystgnr:
Hmm... Nick, remember how my description of this branch was "I remember working on it, and I remember there being something wrong with it that made it unsuitable for merging, but I can't remember what was wrong, and I don't see any errors or omissions?"  My current leading candidate for "what was wrong" is "I forgot that these terms are already implicitly included in the natural boundary condition".

I'm not yet sure about that, though.  I'd have expected the natural BC on the normal velocity to be a Neumann-like "du/dn  = 0", not a Robin-like "du/dn = p/nu".  I'm used to solving INS flows where p is only well-defined up to a constant, for that matter, and that's definitely not the case if changing p changes your BC.  On the other hand, I'm used to solving INS flows with closed and periodic domains...  And on the gripping hand, even though a Robin BC could make sense, I'd expect one with the opposite sign - the higher the pressure is, the more your flow ought to be *decelerating*, not *accelerating* as it approaches the boundary, no?

nicholasmalaya:
The biggest remaining issue to me is: for runs with the current baseline with a free b.c. on 'w' at the top, we are still getting funky high-velocity jets that flow in. That certainly is not physical. My working hypothesis is that this is a result of the pressure being weird on the inflow, which if it was too low, would cause a rapid acceleration downward. This at least fits the results we have seen. However, surely we can make an outflow boundary of this sort sane.

FYI, the b.c. at the top are:
* Natural 'w'
* Parsed u,v,p -- natural if w>0, 0 if w<0. 

(Extra-points for gripping hand reference, btw)

nicholasmalaya:
I finally had a chance to chat with Bob on the phone yesterday and wanted to update this issue. 

I think Bob agrees with Roy strongly on the following, namely:

> I'm not yet sure about that, though. I'd have expected the natural BC on the normal velocity to be a 
> Neumann-like "du/dn = 0", not a Robin-like "du/dn = p/nu".

I believe this is the crux of the issue. Setting the pressure magnitude is not what we want, and is almost certainly what is causing our problems in the simulations. We do want a Neumann-like specification on the pressure at the boundary, in order to specify that there is no pressure gradient at the boundary. 

Now recognizing what we want and how to implement it are admittedly very different things. But I believe the above is what we want to get to.

pbauman:
OK, so it sounds like you don't want the natural outflow, but you want something else. You're now getting beyond my experience. :) I'm forwarding you and Roy some references on the subject. I think whatever you guys decide is appropriate should go into the BCHandler, with an appropriate "outflow" name, but we'll see what happens.

pbauman:
> I'm not yet sure about that, though. I'd have expected the natural BC on the normal velocity to be a Neumann-like "du/dn = 0", not a Robin-like "du/dn = p/nu". I'm used to solving INS flows where p is only well-defined up to a constant, for that matter, and that's definitely not the case if changing p changes your BC.

Just to reply to this - yes, this is exactly the case. If you have a closed domain, then pressure is only defined up to a constant (in fact, in the low Mach case, you have a separate equation for the thermodynamic pressure to solve). If you have a "natural" outflow boundary condition, then the pressure is naturally determined by that boundary condition and there is no null space.

nicholasmalaya:
Just wanted to document that the "pressureInletOutletVelocity" b.c. in openFOAM is described as,

<pre>
Class
    Foam::pressureInletOutletVelocityFvPatchVectorField
Description
    Velocity inlet/outlet boundary condition patches for where the pressure is
    specified.  zero-gradient is applied for outflow (as defined by the flux)
    and for inflow the velocity is obtained from the patch-face normal
    component of the internal-cell value.
</pre>

Essentially, it sounds as if this sets U and p between dirichlet and zeroGradient depending on direction of U. (http://www.openfoam.org/docs/user/boundaries.php)

Not sure this helps us much, but I wanted to at least mention it. Two things jump out as problematic: 
* It still requires a pressure to be specified
* I'm not sure it deals with our 'inflow looks like a jet' problem

nicholasmalaya:
More info, from the header files:

<pre>
Description
    This velocity inlet/outlet boundary condition is applied to pressure
    boundaries where the pressure is specified.  A zero-gradient condtion is
    applied for outflow (as defined by the flux); for inflow, the velocity is
    obtained from the patch-face normal component of the internal-cell value.

    The tangential patch velocity can be optionally specified.

    \heading Patch usage

    \table
        Property     | Description             | Required    | Default value
        phi          | flux field name         | no          | phi
        tangentialVelocity | tangential velocity field | no  |
    \endtable

    Example of the boundary condition specification:
    \verbatim
    myPatch
    {
        type            pressureInletOutletVelocity;
        phi             phi;
        tangentialVelocity uniform (0 0 0);
        value           uniform 0;
    }
    \endverbatim

Note
    Sign conventions:
    - positive flux (out of domain): apply zero-gradient condition
    - negative flux (into of domain): derive from the flux in the patch-normal
      direction
</pre>

pbauman:
Can we close this out?

nicholasmalaya:
I agree, this is a non-issue now and can be closed.

pbauman:
Thanks!

PR #223, opened 2015-02-24
ElasticCable Refactor

pbauman:
Created a base class to be reused by subclasses. In the same vein as ElasticMembrane. CC'ing @cahaynes


pbauman:
Need this right away and it doesn't affect anyone else (that I'm aware of) so I'm merging.

Issue #225, opened 2015-02-25
grins does not honor LIBMESH_LIBS


capitalaslash:
when compiling grins, I had an error in linking phase, where openmpi libraries where not found.
in fact the linking command was missing -L/usr/lib/openmpi (and adding it to LDFLAGS solved it for me), but it is present in the LIBMESH_LIBS variable in the Makefile.

pbauman:
First, thanks for reporting this and thanks for your interest in GRINS! If you're willing, post your interests to grins-users or email me privately so that I make sure we don't interfere with your work. FYI, there's a major overhaul of the input file coming shortly.

To your report, are you trying to specify your own compilation flags at compile time using the `--disable-libmesh-flags ` flags? If not, there may be an issue --- we pull the flags from `libmesh-config`. Could you confirm that `-L/usr/lib/openmpi` is or is not present in 
`$LIBMESH_DIR/bin/libmesh-config --libs`, where `$LIBMESH_DIR` is where you've installed libMesh. By default this reports flags for `METHOD=opt`. You can prefix the call with `METHOD=devel` or `METHOD=dbg` if you built GRINS in either of those modes.

I can also have GRINS echo the included libs next to the currently echoed `$LIBMESH_DIR` and include flags.

capitalaslash:
I'm just doing my first steps with grins to understand if it fits my needs. I'll come up in the mailing list as soon as I know if I will become a /real/ user... I'll wait and see the update on the input file syntax.

I am not using `--disable-libmesh-flag`. my `configure` looks like

```bash
configure --prefix=<> --enable-cxx11 --with-libmesh=<> --enable-antioch --with-antioch=<>
```

the output of `libmesh-config --libs | tr ' ' '\n' | grep openmpi` is

```bash
-L/usr/lib/openmpi
-Wl,-rpath,/usr/lib/openmpi
```

so it seems to be ok. furthermore, `LIBMESH_LIBS` inside the grins Makefile stores the correct value, but it is not used in the linking process. my experience with autotools and libtool internals is very limited and i don't know how to further investigate it.

just to clarify, i'm using latest master versions of antioch, libmesh and grins.

pbauman:
> 'm just doing my first steps with grins to understand if it fits my needs. I'll come up in the mailing list as soon as I know if I will become a /real/ user... 

Sounds good. Please don't hesitate to ask questions.

> I'll wait and see the update on the input file syntax.

I've opened the discussion on grins-devel.

[...]

PR #226, opened 2015-02-25
Use LIBMESH_LIBS and display in configure summary

pbauman:
Found this in working in #225 with @capitalaslash. If the libtool file is not found, we were using LIBMESH_LDFLAGS, but LIBMESH_LIBS is what gets populated by libmesh-config. Also now display LIBMESH_LIBS in the configure summary.

### DIFF
--- a/m4/config_summary.m4
+++ b/m4/config_summary.m4
@@ -34,6 +34,7 @@ echo Library Dependencies:
 echo libMesh....................... : $LIBMESH_PREFIX
 echo libMesh CXXFLAGS.............. : $LIBMESH_CXXFLAGS
 echo libMesh INCLUDE............... : $LIBMESH_INCLUDE
+echo libMesh LIBS.................. : $LIBMESH_LIBS
 #echo QUESO......................... : $QUESO_PREFIX
 #echo Trilinos...................... : $TRILINOS_PREFIX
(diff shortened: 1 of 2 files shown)


roystgnr:
Technically you want to be using both LDFLAGS and LIBS.  The latter is for library names (e.g. -lmesh) and the former is for other compiler flags (e.g. -Lmy_dir).

https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html

This is sometimes important; I've run across linkers that whine and die if you don't give all the LDFLAGS arguments before all the LIBS arguments, for example.

Looks like libmesh-config is giving wrong output.  I'll take a look at it.

pbauman:
That's what I thought, but I was going to for the quick/easy fix. Sorry bout that.

roystgnr:
Ah, now I remember why libmesh-config is lumping everything into LIBS.  Because that's what PETSc is doing, and what SLEPc is doing, and what Trilinos is doing, and okay now it's not worth my time anymore.

Let's still include libMesh LDFLAGS in GRINS, though, just in case I do ever get the hankering to regexp-apart third party linker options.

pbauman:
Replaced by #253. Thanks @roystgnr

PR #227, opened 2015-02-26
Pass physics_name to flow_vars in IncompressibleNavierStokesBase

pbauman:
Contrary to 08e99bc49c27c8fdd6ab86336aa310fa05d39774, this really
should be physics_name because Stokes is subclassing this and
we should be able to parse Stokes section. Uncovered this as part of
turning on UFO detection.

I actually think that the variable classes could actually be treated
as "mixins" and we use multiple inheritance in the root physics classes
with these variable classes. Another day...

Make check is passing for me with this change.


pbauman:
Double checked @vikramvgarg's Spalart-Allmaras branch and this shouldn't affect it, so I'm going ahead and merging.

Issue #178, opened 2014-12-07
Consistency in property input specification


pbauman:
Right now, properties, models, values, etc. are spread out all over the place and this is bad. We need to overhaul this.

My suggestion: Have a `[Properties]` section and each subsection deals with its own thing and all the relevant `Physics` classes use. So `Properties/Viscosity`, `Properties/Conductivity`, `Properties/Chemistry`, etc.

@nicholasmalaya, @roystgnr, other suggestions?

pbauman:
This was completed by #317.

PR #228, opened 2015-02-26
Enable GetPot UFO Detection

pbauman:
[GetPot UFOs](http://getpot.sourceforge.net/node15.html)

This turns on GetPot UFO detection at the end of the `Simulation` constructor (i.e. once everything should've been initialized). I've chosen to error if there are any detected. A list of the unused variables will be printed to `libMesh::err`. 

I've updated all the regression tests, but still need to do the examples. At least wanted to put this up so folks could pull and start preparing their input files for when this gets merged.

First in line for addressing #178.


nicholasmalaya:
image

Thanks Paul.

roystgnr:
:+1:

pbauman:
OK, I've cleaned up all the regression tests and the examples. I wiped out a few of the examples that were turds at this point. There's still a few "in-progress" examples that I left (laminar_flame, mass_injection, simple_prop) that don't work (not because of UFOs...), but we might want to pick at in the future. This was rebased on #229. This is good to go whenever @roystgnr and/or @nicholasmalaya are ready.

Note that `make check` should in theory work all through the history. I updated all the regression test input files and then committed turning on UFO detection.

pbauman:
We good here? I'd like to merge this 1. We need this and 2. I'm proceeding with the input file refactoring and I want to have this merged before I start changing the regression and example input files.

roystgnr:
I thought ":+1:" and "Thanks Paul" were adequate feedback.  I'll get more creative in the future.

pbauman:
It usually is, but since this will break runs, I wanted to be doubly sure. Sorry for being overly pedantic.

roystgnr:
It'll break runs, but unless I misunderstood it'll only break runs in the friendliest way: immediate exit and notification regarding which UFO was the problem.  It's not actually deprecating any working options, just screaming about broken ones.

nicholasmalaya:
Here is a cute issue, let's say you have a surrogate variable being used but only indirectly, e.g.

radius='1.0'
area='pi*${radius}**2'

Then I am seeing radius flagged as an unused variable. I'm guessing catching this case is not as trivial, but this is going to be impossible to work around for us, since several of the vane tiers are sufficiently complex that I would be terrified to set them up as a single penalty function line.

pbauman:
You need to use the full "path", even if it's in the same section

e.g.

[Section]

helper_var

var = '${Section/helper_var}'

nicholasmalaya:
Oh. Got it. Sorry, I must have missed this in the ticket. Sorry to make you baby me through it.

PR #229, opened 2015-02-26
Update IncompressibleNavierStokesBase constructor

pbauman:
Past me was a real @$$hole and didn't explain  08e99bc very well, leading to premature merging of #227 which broke anyone using stabilization and non-Taylor-Hood elements.

Now pass "my_physics_name" and "core_physics_name" to tell the FlowVariables class where to look for the FE inputs in the input file. Also added a convection_cell regression test that would've hit this when I did `make check` with #227. This is smaller version of the convection_cell example.


roystgnr:
Looks much better, thanks!

PR #230, opened 2015-02-27
Refactor mesh options input file specification

pbauman:
See [this](https://groups.google.com/forum/#!topic/grins-devel/AGVvODmfzh8) thread.

If you want to generate the mesh, the new input specification looks like

```
[Mesh]

    class = 'serial'

    [./Generation]
        dimension = '1'
        n_elems_x = '10'
        x_min = '0.0'
        x_max = '1.0'
        element_type = 'EDGE2'
[]
```
etc.

If you want to read the mesh:
```
[Mesh]

      class = 'parallel'

      [./Read]
            filename = 'filename.exo'

[]
```

I need to update the regression tests and examples. I'm going to use this opportunity to make a master example input file with comments for all the options.

Now's your chance to complain if you don't like a section/variable name.

`make check` is passing and spitting out lots of deprecated warnings.

Note the diff is a little ugly because I introduced a helper function that threw off a lot of spacing.

### DIFF
--- a/src/solver/src/mesh_builder.C
+++ b/src/solver/src/mesh_builder.C
@@ -101,13 +101,13 @@ namespace GRINS
 
     // Were we specifically asked to use a ParallelMesh or SerialMesh?
     {
-      std::string mesh_class = input("Mesh/mesh_class", "default");
+      std::string mesh_class = input("Mesh/class", "default");
 
       if( input.have_variable("mesh-options/mesh_class") )
         {
           std::string warning = "WARNING: mesh-options/mesh_class is DEPRECATED.\n";
-          warning += "         Please update to use Mesh/mesh_class.\n";
-          warning += "         Mesh/mesh_class can take values: serial, parallel\n";
+          warning += "         Please update to use Mesh/class.\n";
+          warning += "         Mesh/class can take values: serial, parallel\n";
           grins_warning(warning);
 
           mesh_class = input("mesh-options/mesh_class", "default");
@@ -121,7 +121,7 @@ namespace GRINS
         mesh = new libMesh::Mesh(comm);
       else
         {
-          std::string error = "ERROR: Invalid mesh_class "+mesh_class+" input for Mesh/mesh_class.\n";
+          std::string error = "ERROR: Invalid class "+mesh_class+" input for Mesh/class.\n";
             error += "       Valid choices are: serial, parallel.\n";
             libmesh_error_msg(error);
         }
--- a/src/solver/src/mesh_builder.C
+++ b/src/solver/src/mesh_builder.C
@@ -390,8 +390,9 @@ namespace GRINS
           }
       }
 
-    int uniformly_refine = input("mesh-options/uniformly_refine", 0);
-    
+    int uniformly_refine = input("Mesh/Refinement/uniformly_refine", 0);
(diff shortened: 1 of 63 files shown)


src/solver/src/mesh_builder.C
```diff
@@ -58,144 +61,340 @@ namespace GRINS
...
+    libMesh::Real x_min = input("Mesh/Generation/x_min", 0.0);
+    if( input.have_variable("mesh-options/domain_x1_min") )
+      {
+        std::string warning = "WARNING: mesh-options/domain_x1_min is DEPRECATED.\n";
+          warning += "         Please update to use Mesh/Generation/x_min.\n";
+          grins_warning(warning);
+
+          x_min = input("mesh-options/domain_x1_min", 0.0);
+      }
+
```
roystgnr: You've got about a brazillion of these blocks, which really ought to be factored into a single (template) function.

I'd also upgrade from warning to error in the case where both the new and the old option is found in the same config file.  Otherwise I could imagine new-style input file A accidentally overriding itself by including old-style input file B.

roystgnr:
Looks decent enough with or without my suggested refactoring and paranoia.
:+1:
Thanks, Paul.
Unambiguous Approval

pbauman:
Actually, I now take back my apology for pedantry --- I like the confetti.

pbauman:
> You've got about a brazillion 

Is that like a 'brah'-zillion?

> of these blocks, which really ought to be factored into a single (template) function.

Someday I'll suppress my copy/paste-noob-programmer tendency for things like template functions. I'll cook up something better.

> I'd also upgrade from warning to error in the case where both the new and the old option is found in the same config file. Otherwise I could imagine new-style input file A accidentally overriding itself by including old-style input file B.

I'd initially dismissed this (cause why would you only go half way in getting rid of the annoying messages?), but you're probably right. It's easy enough to add a helper function where we check all those to make it easier to remove later. I'll do that.

pbauman:
Add example input file. Anything I'm missing in the comments so far?

pbauman:
It just occurred to me that the 'type' variable is really unnecessary - we should be able to check if the "Read" or the "Generation" sections are present and proceed accordingly.

Opinions: should we do that or do you think have the 'type' variable adds to the readability? For things like this, I figure we decide here and move forward with that for similar occurrences in subsequent PRs.

roystgnr:
The type variable may add to usability.  It's slightly easier to switch between two different meshes if you can comment out one of two lines rather than one of two sections.

Omitting the type variable is more readable, though.  I'm happy either way.

pbauman:
You'd have to uncomment the other section anyway right? Because

ufo

roystgnr:
Good point.  Let's omit the type then.

pbauman:
Helper function saved ~80 lines. Still need to check for collisions, eliminate Mesh/type option and update tests and examples.

pbauman:
As of bffa8bf, this relies on libMesh/libmesh#481.

pbauman:
Tests are done, just need to update examples.

@roystgnr could you please double check my descriptions of locally_h_refine and redistribution in the examples/master_example_inputfile.in? If they need updating, you can post text here and I'll add/update it. TIA.

pbauman:
Barring any corrections on the example input, this is done!

PR #234, opened 2015-03-03
Continue after max iterations

nicholasmalaya:
This pull request adds a new input line to the parser, permitting the user to dictate that the routine either throws an exception or continues when the non-linear solver does not converge. In particular, in the latest GRINS, the default behaviour is to continue on the merry way, such as,
<pre>
 Nonlinear solver reached maximum step 10, latest evaluated residual
0.291611                                                    
  Continuing...           
</pre>

Now, the default behaviour is to throw an exception, unless the flag, 'continue_after_max_iterations' is set to true. I would prefer this to be the case, as I want to know this is happening and explicitly give permission, but this is admittedly a change from the previous GRINS settings, and I'm happy to switch it back to 'true' (and therefore continuing even without convergence) if that is preferred. 

However, I do feel slightly bad, as this appears to be introducing a warning at compilation, namely:
<pre>
make[2]: Entering directory `/scratch/nick/sov_libs/nick_grins/build/src'
  CXX      solver/src/grins_solver.lo
../../src/solver/include/grins/grins_solver.h: In constructor 'GRINS::Solver::Solver(const GetPot&)':
../../src/solver/include/grins/grins_solver.h:80:10: warning: 'GRINS::Solver::_continue_after_max_iterations' will be initialized after [-Wreorder]
</pre>

Not totally sure what is causing this. Otherwise the feature is working as expected for me.

### DIFF
--- a/src/solver/src/grins_solver.C
+++ b/src/solver/src/grins_solver.C
@@ -42,7 +42,6 @@ namespace GRINS
 
   Solver::Solver( const GetPot& input )
     : _max_nonlinear_iterations( input("linear-nonlinear-solver/max_nonlinear_iterations", 10 ) ),
-      _continue_after_max_iterations( input("linear-nonlinear-solver/continue_after_max_iterations", false ) ),
       _relative_step_tolerance( input("linear-nonlinear-solver/relative_step_tolerance", 1.e-6 ) ),
       _absolute_step_tolerance( input("linear-nonlinear-solver/absolute_step_tolerance", 0.0 ) ),
       _relative_residual_tolerance( input("linear-nonlinear-solver/relative_residual_tolerance", 1.e-15 ) ),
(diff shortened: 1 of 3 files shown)


roystgnr:
You've declared member variables in one order and initializer list in another order.  Since the member variables get initialized in the *first* order the compiler thinks you're confused and warns you.

src/solver/src/grins_solver.C
```diff
@@ -42,6 +42,7 @@ namespace GRINS
 
   Solver::Solver( const GetPot& input )
     : _max_nonlinear_iterations( input("linear-nonlinear-solver/max_nonlinear_iterations", 10 ) ),
+      _continue_after_max_iterations( input("linear-nonlinear-solver/continue_after_max_iterations", false ) ),
```
pbauman: Make this the last one in the initialization list and your warning should go away.

nicholasmalaya:
Thanks guys, I did not know the compiler wanted consistent member variables and initialization. Good to know.

pbauman:
Does `make check` pass? You've set the default to false which may cause a couple of the nastier solids regression tests to die.

libMesh::DiffSolver has the default as true, but I'm wavering on which I think is better.

nicholasmalaya:
Sorry, should have detailed that originally: As remarkable as it sounds, all the tests are passing for me on my ices build.

pbauman:
OK, thanks.

PR #235, opened 2015-03-09
Update input for antioch_kinetics_rates app

pbauman:
Makes more sense to take thermodynamic pressure from input
than the density.

### DIFF
--- a/src/apps/antioch_kinetic_rates.C
+++ b/src/apps/antioch_kinetic_rates.C
@@ -59,9 +59,9 @@ int main(int argc, char* argv[])
   libMesh::Real T0 = input( "Conditions/T0", 300.0 );
   libMesh::Real T1 = input( "Conditions/T1", 300.0 );
   libMesh::Real T_inc = input( "Conditions/T_increment", 100.0 );
-  
 
-  libMesh::Real rho = input( "Conditions/density", 1.0e-3 );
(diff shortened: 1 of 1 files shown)


src/apps/antioch_kinetic_rates.C
```diff
@@ -98,7 +101,9 @@ int main(int argc, char* argv[])
   while( T < T1 )
     { 
       Antioch::TempCache<libMesh::Real> T_cache(T);
-      
+
+      libMesh::Real rho = p0/(R_mix*T);
+
       antioch_kinetics.omega_dot( T_cache, rho, Y, omega_dot );
```
roystgnr: If this is more convenient to you, let's merge it.  But from a "chemistry is tricky" standpoint, I prefer keeping all those calculations stuffed into Antioch over a hard-coded ideal gas law in GRINS.  We're not planning any ludicrously-high-pressure kinetics *now*, sure, but we might later, and it would be easy to miss an equation of state if they're not all in the same place.

pbauman: Agreed. I need to dig into what Sylvain has done - I'm not sure we have methods for such calculations in Antioch yet, but in the short term, this is definitely simpler. Will open a new ticket to capture the thought.

PR #236, opened 2015-03-09
Adaptivity printqoi fix

pbauman:
Only print the QoI if the user requests it in the adaptivity solver.

Really, we need to do additional checking on if the user has actually set a QoI to handle cases were they set print_qoi = 'true', but don't actually have a QoI; will open a ticket for that. For now, this is a quick fix.


roystgnr:
:+1:

PR #241, opened 2015-03-11
Fix Postprocessing Index Heisenbug

pbauman:
Each Physics class is responsible for its own postprocessing variables (following #167) by registering with PostprocessedQuantities and getting an `unsigned int` back so that PostprocessedQuantities can later request the evaluation of that quantity without using string comparisons. There were 2 issues here and I can't believe they didn't bite us earlier in much worse ways.

1. Many of the index variables weren't initialized in the Physics classes so we were getting uncertain behavior. This is why things changed based on the order requested, number of processors, etc.

2. PostprocessedQuantities was counting from zero so if the default for the `unsigned int`s was zero, and the 0 variable was requested, you would've gotten the first quantity that was indexed at zero, whether you requested it or not.

I think this is still a design problem (I'll create an issue for discussion on this point), but this PR fixes the issue by 1. having PostprocessedQuantities count from 1 for the indices referenced in the Physics and 2. Initializing the bloody variables (somewhere Scott Meyers is dying inside).

@nicholasmalaya can you confirm this fixes your conductivity plotting issue? This fixed my plotting issue.

Sorry for the trouble.

### DIFF
--- a/src/physics/src/heat_transfer.C
+++ b/src/physics/src/heat_transfer.C
@@ -44,7 +44,8 @@ namespace GRINS
 
   template<class K>
   HeatTransfer<K>::HeatTransfer( const std::string& physics_name, const GetPot& input )
-    : HeatTransferBase<K>(physics_name, input)
+    : HeatTransferBase<K>(physics_name, input),
+    _k_index(0)
   {
     this->read_input_options(input);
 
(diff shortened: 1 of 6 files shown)


nicholasmalaya:
Trying it now, thank you for the help with this. Will update in a few.

nicholasmalaya:
Builds and appears to function correctly on lonestar! Thanks Paul.

pbauman:
Sweet.

PR #245, opened 2015-03-19
Initialize libMesh earlier

roystgnr:
Otherwise we aren't ready for any libmesh_error() calls from the GetPot initialization.

That was originally an infinite loop prompting https://github.com/libMesh/libmesh/pull/503 but even with that fixed we can't query the libMesh command line for gdb version until after LibMeshInit construction.


pbauman:
Thanks!

Issue #243, opened 2015-03-12
output velpen forcing 


nicholasmalaya:
Hi all,

Right now we can output the normal vectors, but not the forcing, calculated by velocity penalty. This is output as 'vel_penalty_'. It would be great to output the forcing as well. I've got a hacked together version I've been using, so this is not time sensitive. @roystgnr correctly pointed out to me that this is a more general feature that should be officially added to grins and propagated back upstream. 

I'm happy to take a lead on this but I was hoping to get a crash coarse on how one 'officially' adds a variable to available for output. If anyone could point me to a location or commit that did this, I would be happy to poke around with this. I would not expect this to make the 0.6.0 release.

pbauman:
> general feature that should be officially added to grins and propagated back upstream

Upstream = GRINS or upstream = libMesh? If former I understand, if the latter, I don't understand.

> I was hoping to get a crash coarse on how one 'officially' adds a variable to available for output

Everything you need to do should be localized to the Physics subclass from which you want to compute the output. Just append what's there for `VelocityPenalty` already. Namely:

1. Cache an `unsigned int` in the Physics subclass (in this case `VelocityPenalty`) for the local variable number that you get back when registering. For the love of all things holy, initialize it to 0 in the initialization list (see #241).

2. In `VelocityPenalty::register_postprocessing_vars`, check for the variable name that you want to output (whatever you want, that's what will be in the input file to trigger it's output), and register the variable. You pass the string name in to `postprocessing.register_quantity()` and you get back an `unsigned int`, i.e. the cache variable from step 1.

3. In `VelocityPenalty::compute_postprocessed_quantity`, check if the `quantity_index` passed in matches your cached index. If so, do your sauce to compute the quantity and set the variable `value` to the value of your new postprocessed quantity; this is what will be output.

4. Compile and test.

Please let me know if there's anything unclear or if you get stuck. I tried to make this as painless as possible, so I'll be interested if there's any way I can improve it (beyond #175  and #242).

Right now, the postprocessing is only doing a local L2 projection onto a linear basis (there's no input knob for an order). So, if your new quantity is a derivative, it's probably going to be crappy at element boundaries and will be discontinuous across elements. Burning on my todo list is adding gradient-recovery quantities ala Zienkiewicz-Zhu. (Basically averaging derivatives over node patches. This is how stress postprocessing is typically done, for example.)

pbauman:
Looks like this was satisfied awhille back. Ref. discussion in #246.

PR #246, opened 2015-03-19
Plot_forcing

nicholasmalaya:
This update permits outputting the forcing calculated by the velocity penalty. This closely follows the recommendations generously provided by @pbauman in #243. No major deviations. Update is not too large, should be easy to digest.

However, this is a *do not merge* at present. The latest revision is building and working, but fails when using "output_vars = 'velocity_penalty'" in a velocity_penalty section. I think, however, we are close. My only point of confusion occurs in velocity_penalty_base.C in the method 'compute_force' (line 99). I've added a boost autoptr named, 'force_velocity_function' which will be accessed in 'velocity_penalty.C' in compute_postprocessed_quantity (line 255). 

However, @roystgnr 's approach for the other quantities followed, 
<pre>
(*base_velocity_function)(point, time,
                              output_vec);
</pre>
Where, output_vec is a DenseVector, e.g. libMesh::DenseVector<libMesh::Number> output_vec(3);
What I want is to do the same thing for F, our forcing vector. However, this is not in a DenseVector, so,
<pre>
    (*force_velocity_function)(point, time,
                               F);
</pre> isnt working for me.

Any eyes that can be spared are appreciated.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,6 @@
 Version 0.6.0 (In progress)
+   * permit output forcing calculated by velocity penalty
    * Added velocitypenalty2 + velocitypenalty3, to support 3d constraints 
-
    * Added 'continue_after_max_iterations' option to parser
    * Added velocitypenalty2 + velocitypenalty3, to support 3d constraints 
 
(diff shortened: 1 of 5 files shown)


nicholasmalaya:
Killing this pull request: feature is available with the re-factored output variables.

PR #248, opened 2015-03-21
Simulation constructor refactor

pbauman:
Created several helper functions to condense code in the constructor. Created a new constructor that additionally takes a GetPot& for the command line parsing. Deprecated the previous constructor.

This is in preparation for adding command line flag to warning instead of error on UFO detection. I went this route because it looked unsafe to have a single GetPot object parse both the input file and command line. Also didn't like the idea of shoving it in the SimulationBuilder. And it's probably a good idea to have access to the command line anyway.


roystgnr:
Refactoring looks sane.

PR #249, opened 2015-03-21
Enable only warnings for UFO detection with command line option

pbauman:
Following some complaints about the lack of this option, this PR enables only warnings, instead of errors, if UFOs are detection. Namely, add the option "--warn-only-unused-var" on the command line sometime after the input file (the input file must be the first option listed after the grins executable (we should probably add some flexibility here)) and any detected UFOs will only illicit a warning and the list of UFOs instead of an error. I've also added unit tests for both the warning and the error versions.

This builds off of #248 (for command line access in `Simulation`).


nicholasmalaya:
FYI, I am confirming this capability is working for me.

pbauman:
Rockin, thanks.

PR #250, opened 2015-03-21
Echo GRINS and libMesh hashses as well as command line.

pbauman:
Wish I'd done this about 4 months ago (or earlier...). Should help with traceability.


roystgnr:
Thanks!

Issue #247, opened 2015-03-20
Forcing Source


nicholasmalaya:
This issue request is to add a forcing function to GRINS. This permits, when specified by the input file, to read a parsed function that will then spatially force the velocity field in some region. 

This is we can be less restricted in terms of manipulating the flow than with the velpenalty.  In particular, to start, the most useful feature will be to add a vertical forcing in a small region near the ground, inside the boundary layer. 

I chatted with Roy about this in the AM, although I am still not totally sure how to implement this guy. It sounds as if we can use a similar infrastructure to the velocity penalty, but now with a simpler structure given that the forcing does not require normal vectors, etc.

nicholasmalaya:
A few questions: 

A) Do we think this might be more generally useful for GRINS users? 
B) Should I instantiate it as a new piece of physics, al la velocity penalty?
C) Obviously forcing like this is not energy conserving, so is there an easy way for someone to calculate the energy introduced by the forcing? I suppose I would need to be able to integrate over a region of forcing.

pbauman:
> Do we think this might be more generally useful for GRINS users? 

Yes. In fact, I've been talking with @cahaynes about generalizing the input of source terms (there's at least 2 in GRINS right now and they're overly specialized). We ought to have a just a generic `SourceTerm` (or something) with `Constant` and `Parsed` versions that looks for the variables to which you're adding the source term and grabs the value/expression for the source term and just adds it to the residual (FE type, etc. would be grabbed from the Variable info).

I was originally thinking we would be running headlong the "Reusing a single `Physics` class in the same simulation problem", but there's no reason we couldn't just have this thing soak up all the source terms in one class (might be nice to have multiples for modularity, but this would be a start).

I'd need to do the Variables-input-refactoring first.

Thoughts?

> C) Obviously forcing like this is not energy conserving, so is there an easy way for someone to calculate the energy introduced by the forcing? I suppose I would need to be able to integrate over a region of forcing.

This sounds kind of specialized. Not sure how I'd want to handle this, but I'm open to suggestions.

pbauman:
#47 was my first expression of regret along these lines.

nicholasmalaya:
I'm afraid I'm going to need this capability sooner rather than later. I'm guessing this means I should just plan on creating a new physics class (with the associated variable name caching and initing the variable indices, etc.) and that can be reduced down in a refactor at a later date?

[...]

Issue #46, opened 2013-09-13
Refactor HeatTransferSource to VolumeSource


pbauman:
Or some such name. It occurred to me that we can probably make this much more generic by reading the variable names we want to associate the source with and be able to add the source to that residual.

pbauman:
This was taken care of with #251.

PR #251, opened 2015-03-25
Generic Source Functions

pbauman:
Both constant and parsed versions (because FunctionBase is virtual on operator(), was avoiding that for the constant case, probably premature optimization). Have only a weak test for the constant version. This should take take care of #247, #46 (modulo nonlinear source terms).

@nicholasmalaya, please be my guinea pig. Fetch this branch and let me know how it works for you.

Comments welcome.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -121,6 +121,7 @@ libgrins_la_SOURCES += physics/src/elastic_cable_constant_gravity.C
 libgrins_la_SOURCES += physics/src/nonlinear_elasticity_instantiate.C
 libgrins_la_SOURCES += physics/src/source_term_base.C
 libgrins_la_SOURCES += physics/src/constant_source_term.C
+libgrins_la_SOURCES += physics/src/parsed_source_term.C
 
 # src/properties files
 libgrins_la_SOURCES += properties/src/constant_conductivity.C
@@ -314,6 +315,7 @@ include_HEADERS += physics/include/grins/elastic_membrane_constant_pressure.h
 include_HEADERS += physics/include/grins/elastic_cable_constant_gravity.h
 include_HEADERS += physics/include/grins/source_term_base.h
 include_HEADERS += physics/include/grins/constant_source_term.h
+include_HEADERS += physics/include/grins/parsed_source_term.h
 
 # src/properties headers
 include_HEADERS += properties/include/grins/property_types.h
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -119,6 +119,7 @@ libgrins_la_SOURCES += physics/src/elastic_cable_base.C
 libgrins_la_SOURCES += physics/src/elastic_cable.C
 libgrins_la_SOURCES += physics/src/elastic_cable_constant_gravity.C
 libgrins_la_SOURCES += physics/src/nonlinear_elasticity_instantiate.C
+libgrins_la_SOURCES += physics/src/source_function_base.C
 
 # src/properties files
 libgrins_la_SOURCES += properties/src/constant_conductivity.C
@@ -310,6 +311,7 @@ include_HEADERS += physics/include/grins/elastic_cable_base.h
 include_HEADERS += physics/include/grins/elastic_cable.h
 include_HEADERS += physics/include/grins/elastic_membrane_constant_pressure.h
 include_HEADERS += physics/include/grins/elastic_cable_constant_gravity.h
+include_HEADERS += physics/include/grins/source_function_base.h
 
 # src/properties headers
 include_HEADERS += properties/include/grins/property_types.h
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -119,7 +119,7 @@ libgrins_la_SOURCES += physics/src/elastic_cable_base.C
 libgrins_la_SOURCES += physics/src/elastic_cable.C
 libgrins_la_SOURCES += physics/src/elastic_cable_constant_gravity.C
 libgrins_la_SOURCES += physics/src/nonlinear_elasticity_instantiate.C
-libgrins_la_SOURCES += physics/src/source_function_base.C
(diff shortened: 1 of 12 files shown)


nicholasmalaya:
@pbauman you rule, I'm building this now and will report back asap. Much, much appreciated.

nicholasmalaya:
I'm getting a,
<pre>
[10] ../../src/solver/src/mesh_builder.C, line 74, compiled Mar 11 2015 at 17:21:37
ERROR: Must specify either Mesh/Read or Mesh/Generation in input.
</pre>

This only appears when I add something to parse, so the following causes an error:
<pre>
value = '1.0*(z<1)'
</pre> 
But this does not:
<pre>
value = '1.0'
</pre>

Do I need to specify that the parsing function is parsed somewhere other than ParsedSourceTerm in enable physics? 

Finally, I'm wondering if there is an easy way to output the forcing function generated from this? If not, please do not worry about it, but would be nice.

pbauman:
On Wed, Mar 25, 2015 at 11:08 AM, Nicholas Malaya <notifications@github.com>
wrote:

> I'm getting a,
>
> [10] ../../src/solver/src/mesh_builder.C, line 74, compiled Mar 11 2015 at 17:21:37
> ERROR: Must specify either Mesh/Read or Mesh/Generation in input.
>
> This only appears when I add something to parse, so the following causes
> an error:
>
> value = '1.0*(z
>
> Did that get truncated? What's the full string?


> Do I need to specify that the parsing function is parsed somewhere other than ParsedSourceTerm in enable physics?
>
>
No. Just make sure the sectioning is setup right. Have a look at the
thermally_coupled_2d_flow input file in test/input_files. It's only
constant, but the parsed version should be analogous.


> Finally, I'm wondering if there is an easy way to output the forcing function generated from this? If not, please do not worry about it, but would be nice.
>
> Will have to add the code. Can get to this later.

pbauman:
Also, it just occurred to me that the function is for all variables and not per variable. Will fix that later too.

For your current thing, I'm not sure what's wrong.

nicholasmalaya:
> No. Just make sure the sectioning is setup right. Have a look at the
thermally_coupled_2d_flow input file in test/input_files. It's only
constant, but the parsed version should be analogous.

That is what I am using. 

> Also, it just occurred to me that the function is for all variables and not per variable. Will fix that later too.

How do you get at a component of velocity? I was using, 

[./Variables]                                                                                                                                                                                                                                
names = 'w'  

But now I am wondering if I need a three component specification, al la velocity penalty.

pbauman:
> How do you get at a component of velocity?

What you have is right. What I meant was if you wanted to do
```
[./Variables]

name = 'T v'
```

There's no way to give a separate function for each, at the moment. I'll change that later. What you have looks correct. I'm not sure what's going wrong in the parsing.

pbauman:
Make sure you have no whitespace in the function string. Can't think of anything else off the top of my head.

nicholasmalaya:
> Make sure you have no whitespace in the function string. Can't think of anything else off the top of my head.

*#^@*!

Working. Thanks Paul.

pbauman:
OK, you're the second person that's bitten. That's a GetPot limitation(?), not FunctionParser, so we should be able to check it. I'll fix it here in this PR and another PR for the other places ParsedFunction is used (unless someone beats me to it).

nicholasmalaya:
Arg, I foolishly misunderstood what we are trying to do. Ideally would like to be able to make this a function of velocity, as well as space. @roystgnr casually mentioned that in principle this might not be too hard, we just need to expose parsedFEMfunction instead of merely parsedfunction. 

If so, would that imply that we just need to: 
1) build some appropriately named module to set in enable_physics
2) include "libmesh/parsed_fem_function.h"
3) essentially the rest would be identical to parsed_source_term, I believe

Sorry to be misleading, I am acting as a source term of confusion these days.

roystgnr:
I think Nick's summary is correct.  Let's merge this now so he can branch the additional changes off master?

nicholasmalaya:
I think this capability is not working. 

I'm happy to share the input file, but the kicker is that if I set the forcing on the vertical component of forcing to be non-zero, and with no bousinesq buoyancy and zero-boundary conditions on the walls, we never get any velocity. 

In other words, the velocity appears to not be being forced...

<pre>
[Physics/ParsedSourceTerm] 
[./Variables] 
names = 'w'  
FE_types = 'LAGRANGE'  
FE_orders = 'FIRST'
[../Function]   
value = '5.0*(z&#60;1)*(sqrt(x^2+y^2)<0.5)'  
</pre>

pbauman:
OK, I'll exercise this harder. There are several classes that can be removed in favor of this. Did you try slamming a hammer to be sure? Not doubting that it's not working, just want to get an idea.

nicholasmalaya:
I'm playing with some stuff right now. Just ran: 

value = '5.0' and that did not do anything, so I'm suspicious anything is happening. By 'do anything' I mean that all the spatial components of velocity, 'u,v,w' are exactly zero. The flow simply is not moving.

nicholasmalaya:
Ok I have also tried, 'value=100', and am getting no motion. Given the back of the envelope scaling I just wrote out, should *certainly* cause substantial velocity flows.

pbauman:
K, sounds like it's broken. I'll look at it tonight.

nicholasmalaya:
Two things:

1) The examples I gave you were not particularly useful, as if the forcing field is just a gradient of a scalar, the entire forcing can go into the pressure, which might not result in a non-zero velocity field. I still believe the forcing is broken, but I'm suggesting that those test cases might not be too illustrative for debugging purposes. value='100*exp(-z)' would work fine, for instance. 

2) Using the new QoI stuff, can one simply output the forcing field into exodus? I tried output_vars='Function' in Variables, ParsedSourceTerm and Function, but that did not do the trick.

PR #252, opened 2015-03-25
Mpirun tests

roystgnr:
So, this turned out to be a much greater undertaking than I'd anticipated.

It seems to be done now, though:

* Solver tolerances have been tightened to the point where we get highly converged solutions independent of processor count.
* New gold standards have been uploaded for a couple cases where we were previously underconverged.
* Tests which need more sophisticated preconditioners in parallel now use them.  For the worst problems (an axisymmetric case, the low mach formulation cases) that means MUMPS as a sub-preconditioner.  For some of those no level of PETSc ILU was cutting it with more than 4 processors.
* Tests which can run in parallel now do so iff your LIBMESH_RUN environment variable is set.
* Two tests which can't run with 12 processors (due to what appear to be libMesh or GRINS bugs, not convergence issues) are set to run in serial and marked "FIXME" in the scripts.
* The result now passes "make check" on my workstation on 1 through 12 processors.


roystgnr:
Never mind, this isn't quite ready to merge yet.

I wasn't expecting preconditioner fixes for higher processor counts to occasionally break convergence at *lower* counts...

roystgnr:
Okay, I think that's better.  We should be safe to merge.  Or we can wait, and I can see if my new MUMPS-free axisymmetric case preconditioner also works robustly with the low mach cases.  And @nicholasmalaya, we really need to play with SoV preconditioning again.  The effect of some of those reordering options was stunning.

nicholasmalaya:
> The effect of some of those reordering options was stunning.

Exciting! Will send a nice input file your way.

roystgnr:
Nope; the ILU with custom reordering works better for the axisymmetric case, but not so well for the low mach cases.  We should merge this as-is.

roystgnr:
No, *now* we should merge as-is.  "make check" doesn't fail until "-np 16", and the MUMPS requirement is gone.  I'll merge today unless someone screams.

roystgnr:
FYI - "make check" will *fail* with this patch unless some of the autotools stuff is rerun.  "re-bootstrap, re-configure" works, of course, but just "./config.status" in the build directory seems to be sufficient.  It looks like the tests don't have their input files listed as dependencies in test/Makefile.am, so make alone doesn't know some of the input files need to be re-generated through configure when their sources change?

pbauman:
Yeah, haven't looked into doing this, always just ran config.status, but might be worth looking into at some point.

roystgnr:
Hmm... one other catch: "make check" isn't a good idea because it wastes (N-1)/N of your CPU cores while compiling tests one at a time.  But "make -j $N check" is a dubious idea, because it launches N^2 processes at once if you've set LIBMESH_RUN to use all your cores.  Does autotools provide a make target for building all tests without running them?

pbauman:
> Does autotools provide a make target for building all tests without running them?

Not as far as I know. Will need to look into it. Maybe we can create an auxiliary check rule that does -j for building and first pass with GRINS_RUN/LIBMESH_RUN with -np 1 and then does -np whatever? I dunno.

PR #253, opened 2015-03-26
Use LIBMESH_LIBS with LIBMESH_LDFLAGS and display in config summary

roystgnr:
This replaces #226 - same functionality but should be more forwards-compatible.


pbauman:
Sweet, thanks.

PR #254, opened 2015-03-26
Parsed interior QoI

roystgnr:
This gives us arbitrary volume integrals of x,y,z,t,u,etc as QoIs.

Sadly:
* It does *not* give us functions of gradient terms u_x, u_y, etc.  We may be able to fix that upstream in libMesh.
* It does not give us gradients of the function, either.  We may also be able to fix that upstream in libMesh via the FParserAD stuff, but for now any adjoint solve is done via finite differencing solution variables.


nicholasmalaya:
Nice! This is killer.

I'm guessing calculating the flux through a surface interior to the domain is non-trivial?

pbauman:
> I'm guessing calculating the flux through a surface interior to the domain is non-trivial?

Might be doable if we can label sidesets for interior elements. Otherwise, not trivial

roystgnr:
So there's only two problems with using this to calculate a flux integral through an interior surface:

1) It can't calculate fluxes
2) It can't integrate on interior surfaces

But other than that we're good.

pbauman:
I'm fine with this as is if it does what you guys need. Might want to add later subdomain restriction option, but if you don't need it now, merge it.

roystgnr:
More seriously: GRINS users will be able to use the upcoming ParsedBoundaryQoI to integrate on surfaces, and libMesh can label boundary ids on interior sides, or even without labels you could do the usual FParser spatial restriction tricks to ensure that you're within 1e-6 of the sides you want, but either way that only works if your coarse mesh happens to line up perfectly with the manifold you want to integrate on.

The flux problem is a libMesh limitation; I'll need to add support for gradient variables in ParsedFEMFunction before that's an option.

We probably should discuss the subdomain restriction option before merging, since there'd be no way to add it later without either using the "empty set of subdomains means integrate everywhere" idiom that Paul doesn't like or the "break compatibility with old input files" idiom that gets us hate mail.  ;-)

I don't think we strictly *need* subdomain restrictions, since it's easy enough to define a region in fparser code via coordinate tests, but we might want them for efficiency or convenience, and if we want to make specifying them mandatory then we need to put them in from the start.

pbauman:
> the "empty set of subdomains means integrate everywhere" idiom that Paul doesn't like

Stuff like this is always up for discussion. Given the recent hate mail, and just input file goofs in general, I've been thinking/leaning towards making the user specify things versus defaults for things like Physics, QoIs, FE types/order, solver types, etc. and only using defaults for things like solver tolerances, screen verbosity, etc. But I admit it's debatable and would like to hear arguments one way or the other.

> don't think we strictly need subdomain restrictions, since it's easy enough to define a region in fparser code via coordinate tests, but we might want them for efficiency or convenience

I keep forgetting about this. You're right, that's enough; for QoI's I'd bet it wouldn't even show up in profiling. It would be nice if that @pbauman @$$hole had made it easy to add documentation for all this.

roystgnr:
I think defaults are perfectly appropriate for cases where we never expect to change the default.

Which, surprisingly, I think includes this case.  We can get GetPot to distinguish between an unspecified vector and an empty vector, which means we can have the former (default) situation indicate "apply on all subdomains" while the latter situation would still unconfusingly indicate "apply on the empty set of subdomains: nowhere".

So if we can change our minds and add subdomain restrictions later without breaking backwards compatibility, there's no reason to worry about it much now.  I'll merge as is.

We may want to open another issue for discussing which variables should or shouldn't take default values.

PR #255, opened 2015-03-26
Actually test vorticity qoi

roystgnr:
Discovered this while adding the Parsed QoI run case.  It turns out that all this time, make check has been verifying that the Vorticity QoI *runs* but hasn't actually been calling the test harness that verifies the output is correct.

We should probably generalize that test harness to use for other QoI cases too, but for now we can at least turn it on for Vorticity.


pbauman:
Holy crap. Thanks for fixing this.

PR #256, opened 2015-03-30
Allow naming of materials for each `Physics` in input file

pbauman:
DO NOT MERGE. This is not yet complete, but it's at a point where feedback can start to be given.

To this point, I've updated the parsing of viscosity to support reading a `material` variable in the input file, while still preserving backward compatibility. This is only enabled for `IncompressibleNavierStokesBase` subclasses to this point. Other materials variables will follow along the lines I've done here, hence the solicitation of feedback at this point.

Main point of feedback: @roystgnr how do you feel about placing of helper functions in their own namespace instead of an object? Since there was no state, I thought this seemed more suitable, but I'm curious what you think about it.

Other comments welcome.

Should we aim just to do viscosity here and then do a separate PR for each update (conductivity, specific heat, stress-strain, etc.)? If so, then what remains here is to update other classes using viscosity, then update tests and examples. I also plan to add tests that exercise the input conflict detection (which will fail and we'll label as XFAIL).


roystgnr:
The advantage of using a namespace is that you don't pass around a useless *this pointer like you would in a non-static class method.

The disadvantage is that if you ever change your mind and decide you need state, you can't add state without API breakage.

I'm not sure what disadvantages there would be of using a class with only static methods.

pbauman:
Oops, forgot to mention, I'm thinking along the following lines for how the `Materials` section will look:

```
[Materials]
   [./Material1] # Non-reacting material
      [./Viscosity]
         model = 'parsed'
         value = 'sin(x)'
      [../Conductivity]
         model = 'constant'
         value = '1.0'
      [../SpecificHeat]
         model = 'constant'
         value = '2.0'

   [../../Material2] # Reacting material
      [./Chemistry]
          library = 'antioch'
          species = 'N O N2 O2 NO'
          datafile = 'kinetics.xml'
      [../Thermodynamics]
          model = 'antioch_CEA' # or whatever the name is
          datafile = 'thermo.dat' #Default will be Antioch's data
       [../Viscosity]
           model = 'antioch_sutherland'
           datafile = 'mu.dat'
       [../Conductivity]
           model = 'antioch_eucken'
           datafile = 'k.dat'
       [../Diffusivity]
           model = 'constant'
           values = '' # Here we just list numbers for each species
[]
```

pbauman:
> I'm not sure what disadvantages there would be of using a class with only static methods.

Good point, I didn't think about static member functions. Will go that route then.

pbauman:
And here we go. Stupidness of not doing this originally is raising it's ugly head. @roystgnr @nicholasmalaya, an incoming change is going to impact your runs, but I'm hoping in a positive way. 

The `BousinessqBuoyancy` stabilization classes are parsing the viscosity only for the constant case and was defaulting to `1.0`. So, if you were using non-constant viscosity, even `ParsedViscosity`, those stabilization terms weren't seeing it. I'll open a new PR for fixing the instantiation of those terms.

Found this because the convection_cell test test was tripping when I was changing that input file to use viscosity from the materials section.

pbauman:
Also, we have a choice to make. Do we want *all* `Physics` to have a `material` variable or do we want some, e.g. the stabilization classes, to be slave to their "core" physics? I'm leaning toward the latter, but I could go either way.

pbauman:
#259 has the fixes for Boussinesq.

roystgnr:
We *definitely* want all the stabilization classes to be slave to their core physics.  If VelocityPenalty wants to use one material and VelocityPenaltyAdjointStabilization another, that's a bug, and the best way to get rid of bugs is to make them impossible to express in the first place.

Ideally we'd want the "subsidiary" physics to *also* be forced to use the same material as the physics they're attached to, too...  If IncompressibleNavierStokes is using one material, then any other flow-based Physics in the same subdomains should be using the same material.  I'm not sure how best to enforce that, though, without inadvertently disallowing multi-material cases.

pbauman:
> We definitely want all the stabilization classes to be slave to their core physics.

OK, this is setup this way as is, so this should be good. (In fact, this is how I found the Boussinesq bug.)

> Ideally we'd want the "subsidiary" physics to also be forced to use the same material as the physics they're attached to, too...

We sort of have this already, at least with the flow physics. Whatever is the "core" physics is the one from which we grab the material. For what's currently there, it gets funky if you want to use one of your `VelocityPenalty` type things with `Stokes`, then you'd have to add (the now superfluous) `IncompressibleNavierStokes` section. Otherwise, I think most of the subsidiary flow physics will follow this rule. Won't swear to it yet though as I don't think I've updated those tests yet.

pbauman:
Closing in favor of #317

PR #257, opened 2015-03-30
Parsed FEM Source function

nicholasmalaya:
This is an addition to parsed_source term. Marking DO NOT MERGE. All regressions are passing but the core capability is not implemented. 

I've just built a new class entitled, ParsedFEMSource which is a duplicate of ParsedSource. I had not had time to work on this until now, but now a tad stuck. I'm unsure what the proper way to get at 
this class includes: 
<pre>
#include "libmesh/parsed_fem_function.h"
</pre>
instead of, 
#include "libmesh/parsed_function.h"                                                                                                                                                                                                       

However, parsedFEMFunction in libmesh, wants, libMesh::ParsedFEMFunction<Output>::ParsedFEMFunction(const libMesh::System&, const string&, const std::vector<std::basic_string<char> >*, const std::vector<_RealType>*) 

I'm wondering what the correct way to get at the libmesh system information is, e.g. where does this information exist in grins? Any help would be most appreciated, because I am not seeing this sort of data obj lying around, at least when I grep about.

I have to run for a moment so I wanted to log where I am, and see if anyone can point me to a good example of this sort of thing in action. I'm just not familiar enough with the GRINS guts to know, yet.


pbauman:
Ugh. My over-engineering `MultiphysicsSystem::init_data` is now biting me. So, you need access to `MultiphysicsSystem` (really just `System`, but `MultiphysicsSystem` is the subclass) after all the variables are added, but we didn't leave a hook in for stuff like this.

I'll create a PR that adds a hook. Better long term solution might be just to get rid of my over-engineering. Can discuss in forthcoming PR.

pbauman:
#258 has the hook you'll need. What you'll do, once that's merged, is cache the function you parse in the constructor. Then, override the auxiliary_init method and then actually create the `ParsedFEMFunction` using the passed in MultiphysicsSystem and the cached string for the function.

Make sense?

nicholasmalaya:
My apologies: this capability was either added or is not necessary. I think it was added but I'm not finding the particular pull. Regardless, this branch is stale and this ticket should be closed (I would label 'reject' if github permitted it).

PR #258, opened 2015-03-31
`Physics::auxiliary_init`

pbauman:
@nicholasmalaya needs a hook in #257 for init'ing the `ParsedFEMFunction`.

### DIFF
--- a/src/physics/include/grins/physics.h
+++ b/src/physics/include/grins/physics.h
@@ -65,6 +65,7 @@ namespace GRINS
   class NBCContainer;
   class DBCContainer;
   class AssemblyContext;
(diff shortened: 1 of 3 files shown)


roystgnr:
Not sure this is the best solution but it's definitely better than no solution.  :+1:

pbauman:
Is the better solution removing all the intermediary functions and just having an `Physics::init_data`? When I first did that, I was thinking having the different methods would help me remember what needed to be done when creating a new `Physics`. Now it just seems to be more limiting.

pbauman:
Regardless, this is short term solution, so I'll go ahead and merge. Should carry on the conversation on grins-devel.

PR #259, opened 2015-03-31
BousinessqBuoyancy*Stabilization viscosity fixes

pbauman:
These two classes got missed during the template-around-viscosity refactoring. If you're using parsed viscosity along with one of these stabilization classes, you got hit by this. This will change your convergence and probably your solutions, but hopefully for the better since the stabilizing test function and tau will now be consistent with the parsed viscosity function. Note mu isn't in the strong form residual here so if your grid was resolving the convection, then this shouldn't change much.

I went ahead and added a convection cell regression that uses parsed viscosity in the input. We reuse the same data, just parse the constant with ParsedViscosity instead of ConstantViscosity, but should help nonetheless.

Sorry @nicholasmalaya. Would you mind giving this a whirl and see what it does? You'll need to start from scratch, I doubt a restart will converge.


roystgnr:
I'll bet a restart will converge.  Didn't it take us forever to be sure of the *sign* error in the Boussinesq stabilization?  Our important runs are usually on a fine enough grid that screwing up subsidiary physics' stabilization is only subtlely wrong.

roystgnr:
Thanks so much for catching this.  Diffs look good to me but I'd like to wait and give Nick a chance to play with it before we merge.

nicholasmalaya:
I did see this ticket. The test run is sitting in the queue on lonestar. I'll give an update when it finally runs.

nicholasmalaya:
Ok, everything is checking out. Does not appear to be making a big difference, likely because we were resolving the convective scales inside the vanes. I would merge this fellow.

PR #261, opened 2015-04-01
Remove turd from HeatConduction

pbauman:
This crap somehow snuck in. Removed.


pbauman:
This is a trivial fix and `make check` passes (testing for this Physics will be coming shortly with point value QoI), so I'm going ahead and merging.

PR #263, opened 2015-04-01
Use the Viscosity model in stabilization terms

roystgnr:
@pbauman caught this in the Boussinesq stabilization but it's also a
problem in the more obscure physics.


pbauman:
:+1:

Whenever you're ready. I don't want to merge in case you want to do a test run first.

PR #264, opened 2015-04-01
SourceTerms Fix + generic solution regression + Laplace regression test

pbauman:
This fixes what @nicholasmalaya mentioned in #251. Sources terms are working now. 

Second, after learning about GetPot being able to parse the command line a few weeks ago, it occurred to me that we can make a generic regression test for all cases where we just want to compare to a gold standard, so I did it. (A future PR will remove all the now redundant special cases.) Hopefully this will dramatically ease the burden of at least adding regression tests. In short, you now add at the command line: input file, gold data file, variables you want to compare, and the norms.

It just occurred to me tol should've been a command line option. Will fix that before merging.

Finally, I used this new thing to make a Laplace regression test that 1. Uses the ParsedSourceTerm and 2. Actually adds a test for HeatConduction that would've caught the BS in #261. I'll build on this test in a forthcoming PR for point value QoIs.


nicholasmalaya:
I confirm that the source terms appear to be working. Before we had no velocity changes, now we have a clear forcing in the field in the region proscribed:

forcing

pbauman:
OK, now I think this is good to go. Will merge later this afternoon (say 4-5 PM EDT) unless other comments/issues arise with this.

pbauman:
Onward and upward!

PR #267, opened 2015-04-02
Output solution sensitivities

roystgnr:
Optional, and defaults to off, of course.

Right now this only works in cases where we're already doing a forward sensitivity solve; we'll probably want it to work in other cases as well later on.


pbauman:
Should we try to warn/error if the user asks for solution sensitivities but is only computing adjoint-based sensitivities?

roystgnr:
Good idea.

pbauman:
:+1:

PR #268, opened 2015-04-02
Catch "print_qoi with no QoIs" error earlier

roystgnr:
We just got a segfault (opt) or a cast failure (devel) before.

Tested and working.


pbauman:
Thanks for the fix.

PR #271, opened 2015-04-02
Output residual sensitivities

roystgnr:
dR/dp should be useful for debugging.

### DIFF
--- a/src/solver/src/simulation.C
+++ b/src/solver/src/simulation.C
@@ -58,6 +58,7 @@ namespace GRINS
     _print_scalars( input("screen-options/print_scalars", false ) ),
     _output_vis( input("vis-options/output_vis", false ) ),
     _output_residual( input( "vis-options/output_residual", false ) ),
+    _output_residual_sensitivities( input( "vis-options/output_residual_sensitivities", false ) ),
     _output_solution_sensitivities( input( "vis-options/output_solution_sensitivities", false ) ),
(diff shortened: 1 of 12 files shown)


src/solver/src/simulation.C
```diff
@@ -309,6 +312,16 @@ namespace GRINS
     context.error_estimator = _error_estimator;
     context.print_qoi = _print_qoi;
 
+    if (_output_residual_sensitivities &&
+        !_forward_parameters.parameter_vector.size())
```
pbauman: Should be useful for adjoint sensitivities too?

roystgnr: Huh... yes, but not right away.  For some reason my code in libMesh is only creating permanent dR/dp vectors (well, retained until the user wants to get rid of them, anyway) in the forward case (and in some Hessian cases), but temporary vectors (deleted before the end of the calculation) in the adjoint case.  I'll need to refactor that before we can get at them.

pbauman:
Cool. :+1:

roystgnr:
Hmm... we're creating temporary dR/dp vectors in the adjoint sensitivity case because we sometimes need those vectors *unconstrained* to properly weight against heterogeneous BC adjoint solutions.  Need to think about this more and set up some better regression tests before refactoring, I think.

pbauman:
Cool. Not urgent, IMO.

PR #273, opened 2015-04-03
Adjoint plotting

pbauman:
Added functions to output adjoint solutions. Have only tested the steady case using the cavity benchmark example. Seems to be doing the right thing.


pbauman:
Probably collides with #271. Go ahead and merge that and I"ll rebase.

roystgnr:
:+1:

roystgnr:
We should add a call to this from the GRINS adjoint_qoi_parameter_sensitivity methods.

pbauman:
Good, suggestion, I'll do that.

pbauman:
Rebased.

pbauman:
I'll go ahead and merge this now. The adjoint solve and management of when to output the adjoint needs an overhaul, which I'll do right now in another PR. At least can get this in if folks want to do some dirty hacks in the mean time.

PR #274, opened 2015-04-03
Adjoint solve refactor

pbauman:
User can now request to do an adjoint solve instead of relying only on using the mesh adaptive solver; this is for steady cases only, no unsteady support yet for adjoint solves. This assumes a QoI has been set, otherwise you'll get an error. We still check the error estimator type in the adaptive solver, so it's enough to have `adjoint_residual` or `adjoint_refinement` error estimator to trigger the adjoint solve. Optionally solve the adjoint problem with
```
[linear-nonlinear-solver]
   do_adjoint_solve = 'true'
```

If you want to output the adjoint solution
```
[vis-options]
   output_adjoint='true'
```
You should be an error if there are no QoI's or if the adjoint problem has not been set to solve.

Note that there was option specific to the mesh adaptive solver for outputting the adjoint solution, but I'm pretty sure I was the only person that new about. None of our examples or tests were using it. So I just removed it without deprecating it. This PR has more general functionality and still supports outputting the adjoint solution.


src/solver/src/simulation.C
```diff
@@ -487,12 +497,57 @@ namespace GRINS
...
+  {
+    // Check if we're doing an adjoint solve
+    _do_adjoint_solve = this->check_for_adjoint_solve(input);
+
+    const libMesh::DifferentiableQoI* raw_qoi = _multiphysics_system->get_qoi();
+    const CompositeQoI* qoi = dynamic_cast<const CompositeQoI*>( raw_qoi );
+
+    // If we are trying to do an adjoint solve without a QoI, that's an error
+    if( _do_adjoint_solve && qoi->n_qois() == 0 )
+      {
```
roystgnr: I think this will fail.  Right now if we don't have any QoIs we don't even bother building a CompositeQoI and then the dynamic_cast will return NULL.

roystgnr:
Everything else looks reasonable at first glance.

src/solver/src/simulation.C
roystgnr: To clarify, by "fail" I mean "segfault rather than die with a nice error message", not anything catastrophic.

pbauman: > I think this will fail. Right now if we don't have any QoIs we don't even bother building a CompositeQoI and then the dynamic_cast will return NULL.

Spot on, just tried it. Good catch.

pbauman:
Pushed a fix that prints an error message and doesn't segfault. Will wait for @roystgnr to approve the fix, then will merge.

PR #276, opened 2015-04-07
Parsed velocity source

roystgnr:
@nicholasmalaya this should be ready to test if you want to give it a try.

@pbauman I'm aware this is a copy-and-pasted hack but it took ~20 minutes to code up, and we need source terms that don't break in the presence of stabilization right away.


nicholasmalaya:
On it, thanks @roystgnr

pbauman:
> I'm aware this is a copy-and-pasted hack

So you're just matching my style...

But seriously, I agree with your earlier comment by email, this is kind of the last straw. I'd like to fix this before the paper deadline, but whenever we are are all past the "needed results 6 months ago" stage, we will do this right.

roystgnr:
On that tangent: previously IIRC we'd decided that "doing this right" meant adding a strong_residual() form in ParticularPhysics for R_i(u), and then each ParticularPhysicsAdjointStabilization would be able to add its own tau*(R(u)*adjoint(dR_i(u))(v)).

I'm starting to lean to instead adding *both* a strong_residual() term and a strong_jacobian() term, so that we wouldn't need ParticularPhysicsAdjointStabilization (or rather we would, but it would just be the existing code, deprecated and retained for backward compatibility.  We would have a single AdjointStabilization class that evaluates tau*(R(u)*adjoint(dR(u))(v)).

nicholasmalaya:
I cannot find any indications that this is malfunctioning. I've run tests, checked the divergence and spot-checked everything. I'm inclined to recommend inclusion of this into GRINS master.

PR #277, opened 2015-04-08
Cable force calculation for elastic_cable physics

cahaynes:
@pbauman


src/physics/src/elastic_cable.C
```diff
@@ -91,33 +91,36 @@ namespace GRINS
...
             if( name == std::string("stress") )
               {
-                // sigma_xx, sigma_xy, sigma_yy, sigma_yx = sigma_xy
-                // sigma_zz = 0 by assumption of this Physics
-                _stress_indices.resize(3);
+                // sigma_xx
+                // sigma_yy=sigma_zz = 0 by assumption of this Physics
+                _stress_indices.resize(1);
 
                 this->_stress_indices[0] = postprocessing.register_quantity("stress_xx");
```
pbauman: We need to differentiate the name between the membrane version is well. `cable_stress`?

src/physics/src/elastic_cable.C
```diff
@@ -91,33 +91,36 @@ namespace GRINS
...
                 this->_strain_indices[0] = postprocessing.register_quantity("strain_xx");
 
-                this->_strain_indices[1] = postprocessing.register_quantity("strain_xy");
+              }
+            else if( name == std::string("force") )
+              {
+                // force_x
+                _force_indices.resize(1);
+
+                this->_force_indices[0] = postprocessing.register_quantity("cable_force_x");
```
pbauman: Drop the `_x` because you're doing the magnitude of the force in the direction locally normal to the cutting plane, right?

src/physics/src/elastic_cable.C
```diff
@@ -358,6 +377,31 @@ namespace GRINS
...
+              {
+                libmesh_error();
+              }
+          }
+        if( is_force )
+          {
+
+            if( _force_indices[0] == quantity_index )
+              {
+                value = tau(0,0)/std::sqrt(I3)*_A;
```
pbauman: We need to add a comment that this is force in the deformed configuration. This won't make a difference in the small strain case (or much of one anyway), but for large strains, this would be important.

pbauman:
Other than those comments, I would like you to rebase instead of merge. So:
1. Create a new branch starting from hash  1566d48 (git checkout  1566d48; git branch cable-force-rebase; git checkout cable-force-rebase)
2. git pull --rebase upstream master. Resolve conflicts along the way. The way to think about that is imagine that you were using the file from master to add your changes - how would you add your changes then? (That's effectively what you'll be doing.)
3. Now commit as usual, push the new branch to your fork, and create a new PR.

src/physics/src/elastic_cable.C
cahaynes: Correct.  Also applies to the stress/strain due to the assumptions.

cahaynes:
Superseded by #279

PR #280, opened 2015-04-17
Another missing RLMNS axisymmetric term + axisymmetric RLMNS regression test

pbauman:
Was missing a mu factor in the momentum axisymmetric term. Also added regression test. Would've picked this up pretty quick had this regression test be in there before... Solution looks OK. MASA test would be better. On my stabilization branch, the unsteady, stabilized, axisymmetric solutions are looking reasonable too, so I'm gaining confidence that this is the last of the terms.


pbauman:
This fix is critical to my stabilization branch, and includes a new test, so I'm going to merge now. I did test with LIBMESH_RUN set to 1 and 4 processors so I think the tolerance should be OK. Please scream if I break you.

pbauman:
To be clear, this only touches the `ReactingLowMachNavierStokes` physics.

PR #282, opened 2015-04-22
More consistent var index initialization

pbauman:
In some cases we initializing variable indices, in some cases we weren't. This is problematic for development cases like forgetting to initialize variables with libMesh since the `VariableIndex` is an `unsigned int` and can often default to 0. This PR defines `GRINS::invalid_var_index` and, for the cases where we've wrapped up the variable indices, use this to initialize the var indices. Note I'm using `std::numeric_limits<unsigned int>::max()` as the invalid value since we can't use negative values. If there's a better suggestion, I welcome it.

### DIFF
--- a/src/physics/include/grins/var_typedefs.h
+++ b/src/physics/include/grins/var_typedefs.h
@@ -28,6 +28,7 @@
 // C++
 #include <string>
 #include <map>
+#include <limits>
 #include "boost/tr1/memory.hpp"
 
 // libMesh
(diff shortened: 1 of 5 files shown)


roystgnr:
@vikramvgarg mentioned today running into some index initialization bug in his code; is this related to that?

pbauman:
Yeah, based on his comment in IM, this should now trigger an assert for him (and in the future for anyone who forget to init these variable containers).

roystgnr:
UINT_MAX is also what we use as libMesh::invalid_uint.

roystgnr:
Looks great; thanks!

pbauman:
>  libMesh::invalid_uint

Damn. I looked for things like invalid_id and such. Didn't think to look for that.

Issue #283, opened 2015-05-08
tag and bag a release


nicholasmalaya:
We have reached an equilibrium in terms of SoV GRINS development. I'm inclined to recommend we tag a release so we have this one for longevity. 

It appears to me that the appropriate hash is around: https://github.com/grinsfem/grins/commit/d4ca6f31898661f113705ecd6499263b32ab6668

which was on Apr 20, 2015. That includes everything (I believe) that I am presently using, and we should be good with this for another month or so, possibly longer.

pbauman:
This seems reasonable to me. I'm kind of deep in a stack that I'll (hopefully...) be unwinding shortly, so if you guys are running stable, this is a good as place as any to tag.

Just for my reference, for updating change log, need to look back to 1a6c07be073a02c47c4f922a19b514b5d79c6bd9

nicholasmalaya:
I don't think I am capable (in terms of permission) to tag a release on github-- anything I can do to help set this up to occur? 

I am not in a rush, but I would like ensure this does happen, since it is an important milestone to have archived long term.

pbauman:
Apologies @nicholasmalaya. I'm under a deadline - once I've saturated the supercomputer queue, I'll tag a release. Either today or tomorrow.

pbauman:
One helpful thing, if you've got nothing better to do, is start identifying existing warnings to clean up and hitting `make distcheck`.

PR #284, opened 2015-05-21
SplitString refactor

pbauman:
The SplitString/int-unsigned int comparison warning has annoyed me to the point of doing something about it. In preparation for the 0.6.0 release (#283).

I just rewrote it because what was there was much longer and less readable than it needed to be. I added a unit test; I srcgrep'd for all the use cases and believe I'm testing them.. Also put the string utilities in their own namespace to facilitate seeing them in the Doxygen documentation (I didn't know we had a `string_to_T` function!).

This should close/supersede #109.

I think `make check` hits all the spots that use this, but just in case, @nicholasmalaya hit this real quick with a case you have lying around?


roystgnr:
Diffs look good.  I'm happy when Nick is.

nicholasmalaya:
I'm officially happy: I appear to be getting sane behavior with this guy.

pbauman:
Thanks!

PR #285, opened 2015-05-21
0.6.0 prep

### LINKED ISSUE #283: tag and bag a release

pbauman:
Builds off of #284.

Removed a few more unused variable warnings. Now the only warnings that are emitted should be from new Antioch and Cantera. `make distcheck` passes for me. Updated CHANGES.

Chime in now if there's anything else you want to try and squeeze in. Otherwise I'll tag and release tonight.

Closes #283.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,8 +1,31 @@
 Version 0.6.0 (In progress)
-   * Added velocitypenalty2 + velocitypenalty3, to support 3d constraints 
-
+   * Refactored SplitString function and renamed split_string.
+     String utilities now in StringUtilities namespace.
+   * Bugfix for variable index initialization. Now if a variable
+     doesn't get properly initialized, an assert should be thrown.
+   * Bugfix for axisymmetric case of ReactingLowMachNavierStokes
+   * New SpalartAllmaras turbulence Physics (thanks to Vikram Garg).
+   * New ParameterManager capability in development. In conjunction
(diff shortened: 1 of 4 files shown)


roystgnr:
If we're releasing master as 0.6.0, could we tag and upload d4ca6f3 as 0.5.9?  The whole point of #283 was to get a convenient tag for backtracking any future regressions, something that hits the version we've been testing heavily *without* additional code we haven't.

pbauman:
> could we tag and upload d4ca6f3 as 0.5.9?

How about 0.6.0-pre?

roystgnr:
0.6.0-pre sounds good.  Looking over the changelog I'm not seeing any actually *scary* code or significant differences; I just want to be paranoid.  While I'm looking at it... let's tag b3dc12f, one commit later than the one Nick's been testing.  It makes no semantic changes but it fixes a compatibility failure with the newest libMesh master.

pbauman:
Rebased. Running `make dist-check` one last time. Then will merge and tag a release.

PR #286, opened 2015-05-22
ParsedBoundaryQoI

roystgnr:
What it says on the box.  For problems without negligible diffusion, this will be more useful when combined with libMesh after https://github.com/libMesh/libmesh/pull/571 so that gradient terms are supported.


pbauman:
Looks good to me. If finite differences become an issue I have an idea to get derivatives in the interim before trying to plug into FParserAD

PR #287, opened 2015-05-22
Fix Nusselt number for the overlapping bc_ids case

roystgnr:
I noticed this when working on the ParsedBoundaryQoI code.  libMesh
boundaries can have multiple ids, and we might specify those ids in
the qoi, and if we do then we probably don't want to double-count
contributions from the overlapping boundary region.

### DIFF
--- a/src/qoi/src/average_nusselt_number.C
+++ b/src/qoi/src/average_nusselt_number.C
@@ -108,41 +108,42 @@ namespace GRINS
   void AverageNusseltNumber::side_qoi( AssemblyContext& context,
                                        const unsigned int qoi_index )
   {
+    bool on_correct_side = false;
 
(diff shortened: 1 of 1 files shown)


pbauman:
After 3 beers, it took me a second to realize the difference between my broken code and your fix. :P Thanks!

I'd like to merge now, before the 0.6.0 release since this is a bug fix. In agreement?

roystgnr:
We've got a real gold standard in low_mach_cavity_benchmark_regression.C, and that's still passing, so yeah, looks 0.6.0-worthy to me.  Otherwise I'd have said no, on the theory that the odds of someone triggering that corner case weren't as high as the odds of me unknowingly introducing a new bug... ;-)

pbauman:
Agreed. I should've explicitly said that since we had the cavity benchmark test that was passing, we're good. Thanks again!

PR #288, opened 2015-05-27
Print header information earlier

roystgnr:
Otherwise a failure in GetPot initialization just goes straight to a stack trace without outputting anything else useful.


roystgnr:
Thanks to @vikramvgarg for catching this.

pbauman:
:+1:

PR #291, opened 2015-05-29
Antioch parameter access

roystgnr:
I haven't added tests yet, but we're compiling (with the newest libMesh master and Antioch!), we haven't broken any regressions, and it ought to be ready for @variscarey to play with too.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -193,6 +193,7 @@ libgrins_la_SOURCES += utilities/src/input_utils.C
 libgrins_la_SOURCES += utilities/src/cached_values.C
 libgrins_la_SOURCES += utilities/src/distance_function.C
 libgrins_la_SOURCES += utilities/src/string_utils.C
+libgrins_la_SOURCES += utilities/src/parameter_antioch_reset.C
 
 # src/visualization files
 libgrins_la_SOURCES += visualization/src/steady_visualization.C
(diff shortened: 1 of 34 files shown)


pbauman:
So, this is the "base" level functionality for users to write code around, or am I missing something where now the user also identifies parameters in the input file similar to previous work?

roystgnr:
You're missing it, yeah; see AntiochMixture::register_parameter().

The user specifies "Antioch/Anything/Sylvain/defined" in a list of parameters, and GRINS will call the Antioch setter/getter methods when it needs to take finite differences using that parameter.

pbauman:
Ah, I see, Ok, I missed that `ParameterAntiochReset` subclassed `libMesh::ParameterAccessor`. Would be nice to have an example, but since it's probably mostly me or @variscarey driving that train, if you're ready for this to go in, let's merge and we'll drop an example in later.

roystgnr:
Not ready for this to go in without testing.  Since @variscarey is probably still settling in, @hovr2pi might take first crack.

pbauman:
For reference, the format is briefly discussed in libantioch/antioch#135. I think I get it now. It will be good to have a test/example to run this through since I'm just about to dealing with the reacting classes for the materials-input-parsing-refactor (see my `input-materials-redo` branch).

roystgnr:
Actually, I would like to merge just 64cfa07e3aca678b655462904c422703e6c15f94 and then rebase the rest of this PR on master.  I've got some other upgrades coming and there would be no way to avoid conflicts with the API change.

pbauman:
> Actually, I would like to merge just 64cfa07 and then rebase the rest of this PR on master.

:+1:

roystgnr:
Conflicts resolved in #391 - I'll keep this branch around until that's ready to merge, though, in case the rebasing screwed something up.

PR #295, opened 2015-08-21
Antioch 'KineticsTheory' species transport models

pbauman:
This PR adds the 'KineticsTheory' species transport models from Antioch, including a 2-D reacting flow regression test that uses these models. 

I also renamed the 'AntiochWilke' property classes to better conform to the naming convention (I'm OK with this because it's *very* unlikely anyone has subclassed them anywhere, but if you have, scream now). I also updated the input options, preserving backward compatibility. Before, the option was `Physics/Antioch/mixture_model = 'wilke'`. Now, it is `Physics/Antioch/transport_model = 'mixture_averaged'`. Species model specification is still the same, except now we support 'kinetics_theory' for the models, assuming Antioch was built with GSL (we error out if the user asks for the models and Antioch was not built with GSL).

Right now, we're still just using the default data from Antioch. A future PR will generalize this capability.

I'd like to merge this soon to rebase the reacting-stab-rebase branch on this.


pbauman:
Oops, I pushed one last commit: I'd forgotten to update the test input files to change `mixing_model` to `transport_model`. I've tested (`make check`) all the combinations for supporting backward compatibility and the new versions and all passed.

Should be good to go now, barring any complaints.

pbauman:
Going ahead and merging now.

PR #296, opened 2015-08-27
Spalart Allmaras changes, and pressure pinning set to 1 in test input…

vikramvgarg:
Spalart Allmaras parameter registration, created sa_airfoil example and set pressure pinning value to 1 in examples


pbauman:
Oh my. We need to split this up. 

I think there's at least 5 separate PRs here (with multiple commits within each one): SA parameter registration, internal SA work adding the `infinite_distance` bit (whatever that is... a separate commit explaining that would've been useful and code documentation more so), the airfoil example, changing the pressure pin value to 1, and the preconditioner reuse for the adjoint. Feel free to take a crack at splitting those up. It'll be a couple of days before I can, most likely.

roystgnr:
The kitchen-sink commit is pretty awful, I admit, but there's only a couple hundred lines of changes here, and dependencies between them.  Would ~5 separate commits in a single PR be satisfactory?

pbauman:
Sure, that works.

PR #297, opened 2015-08-28
Update README, Default Solver Options for Backward Facing Step Example

pbauman:
Feedback from Martin Luethi on things that should've been more clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,11 @@ GRINS uses an Autotools build system, so typical GNU build commands are used.
 4. make check (note parallel-tests are supported)
 5. make install
 
+LD_LIBRARY_PATH
+---------------
+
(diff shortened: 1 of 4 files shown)


roystgnr:
This seems strange.  Examining "chkrpath -l grins" I see the path for libmesh; examining "chkrpath -l libmesh_opt.so" I see the path for PETSc.  I'd have thought that recursive definition would have been sufficient for the linker to find PETSc when grins is run.

And wait... it *is* sufficient for me.  I use modules on Sysnet but I only use PETSC_DIR (which the linker ignores) on my laptop; there's nothing in LD_LIBRARY_PATH.  I wonder what's happening differently on Martin's system.

pbauman:
I'd like to open a separate issue to track that and merge this right away --- OK with you @roystgnr?

roystgnr:
Definitely.

PR #302, opened 2015-09-04
Added API for setting require_residual_reduction variable in NewtonSo…

vikramvgarg:
…lver


roystgnr:
No way.  Deleting the user's diff_solver so that you can set one parameter on the new replacement you created?

Let's try again, this time in Solver::set_solver_options() with the other parameters.

PR #306, opened 2015-09-16
Hand-tweaked preconditioner choice, looser regression tolerance

roystgnr:
This works on up to 26 processors for me.  Nearly anything else I try
(including our previous go-to) breaks on 12 or less.


pbauman:
Huh. I thought I'd tried this on 4. Sorry, and thanks.

roystgnr:
This problem is *crazy* sensitive.  I wouldn't be surprised at all if it had worked on 4 processors with the default METIS partitioning but failed on my --enable-parmesh ParMETIS setup.

PR #307, opened 2015-09-16
Check and Error for Invalid Pressure Pinning Coordinate

### LINKED ISSUE #244: Pressure pinning *usually* only works near domain sides
Because pin_value is called from side_constraint (in all physics except plain Stokes), and because by default FEMSystem only integrates on sides which have no neighbors, pin_value is only called on elements which border the domain boundary.

This is probably a fine restriction to have.  It's what I do in fem_system_ex1.  contains_point() is a little costly, so better to call it O(N^{(d-1)/d}) times than O(N) times.

This is a scary restriction to have, though.  I just wasted an hour or two trying to figure out why I wasn't seeing pressure pinning in a run.  I'm not sure how to get the code to emit a warning in such cases, but that might be worth looking into.

pbauman:
Moves the pinning to `element_constraint` from `side_constraint`. The check, in `PressurePinning::check_pin_value`, is to simply do a loop over the active local elements to make sure the `_pin_location` is somewhere in there and throw an error if it's not. I added (XFAIL) tests to check for this. Also, this fix uncovered 3 regression tests that erroneously pinned the pressure; two required new gold standards, the third was silently not pinning the pressure (before the fix).

Should resolve #244

Sorry for the wasted time @roystgnr @vikramvgarg. This was an embarrassing lack of defensive coding from the early days.


pbauman:
Oh, forgot to mention I tried both serial and with `$LIBMESH_RUN="mpiexec -np 3"`. Only with `SerialMesh` though.

src/boundary_conditions/include/grins/pressure_pinning.h
```diff
@@ -56,6 +57,13 @@ namespace GRINS
 		     const std::string& physics_name );
     ~PressurePinning();
 
+    //! Check the mesh to ensure pin location is found
+    /*! We are assuming here that the pin location is on the boundary
+        so we only check the boundary. If the pin location is found, we
+        set _pin_location_found to true (for assertion later). If not found,
```
roystgnr: You say you're only checking the boundary

src/boundary_conditions/src/pressure_pinning.C
```diff
@@ -66,11 +69,43 @@ namespace GRINS
     return;
   }
 
-  void PressurePinning::pin_value( libMesh::DiffContext &context, 
+  void PressurePinning::check_pin_location( const libMesh::MeshBase& mesh )
+  {
+    libMesh::MeshBase::const_element_iterator el = mesh.active_local_elements_begin();
+    const libMesh::MeshBase::const_element_iterator end_el = mesh.active_local_elements_end();
+
```
roystgnr: But then you're iterating over all active local elements without a boundary test?

src/boundary_conditions/include/grins/pressure_pinning.h
pbauman: Oh crap, thanks. I wrote that before the code, realized we wanted to do element checks and then forgot to update the comment.

pbauman:
Fixed comment, thanks.

roystgnr:
*Do* we want to do interior element checks?  A contains_point() test isn't cheap, and it would be nice to continue to do that only ~N^((d-1)/d) times instead of N times.

roystgnr:
The problem before wasn't that users aren't capable of figuring out the location of a boundary point, the problem was that when we don't realize the need, we're not capable of noticing the mistake without automated assistance.  The new member variable here is enough to fix that.

pbauman:
I guess I got confused by the discussion in #244. OK, I'll redo this to use boundary-only/`side_constraint`. The check should only happen with a full `reinit`, but that will add up with lots of AMR.

One question though: if we're going to be only calling `pin_value` from `side_constraint`, do we need to be getting the element FE object or the side FE object when computing the pinning residual? I would've thought the side FE, but in the FEMSystem [example](https://github.com/libMesh/libmesh/blob/master/examples/fem_system/fem_system_ex1/naviersystem.C#L467), you grab the element FE object.

roystgnr:
Wait... are you caching the result of the check?  That would be fantastic!  Previously we were doing a contains_point() test for every element in every residual evaluation.

roystgnr:
Grabbing either the element FE or the side FE should work.

pbauman:
Actually, I only got halfway there --- I cache whether or not we found the point, but don't store the information to use when actually computing `pin_value`. So maybe that's what we should do then: cache the element id (which should get updated by the reinit call if we do AMR) and then just check whether the current element has that id in `pin_value`. Thoughts on that?

pbauman:
It hadn't occurred to me that `contains_point` was not cheap.

roystgnr:
That sounds good, although I'd do a contains_point() call in the pinned point's cached element, just as a sanity check so we don't cut loose again as soon as someone starts going wild with ALE or other moving mesh tricks.

And yeah; contains_point on non-affine elements (which includes even slightly distorted QUAD4 and HEX8) is a whole dim x dim Newton iteration.  Not expensive individually, but it's noticeable when you do it on every element.

pbauman:
OK, updated now to cache the element id and use that to both error for invalid pin location and check if to apply the pin value.

src/boundary_conditions/src/pressure_pinning.C
```diff
@@ -66,16 +69,57 @@ namespace GRINS
     return;
   }
 
-  void PressurePinning::pin_value( libMesh::DiffContext &context, 
+  void PressurePinning::check_pin_location( const libMesh::MeshBase& mesh )
+  {
```
roystgnr: You need to reset _pinned_elem_id to invalid_id here, no?  Otherwise this class will work the first time it's used but may fail after AMR.

pbauman: Damn, you're right. Thanks for catching that.

roystgnr:
:+1:

PR #309, opened 2015-09-19
Remove redundant protected member from HeatTransferStabBase

pbauman:
Came across this in the `input-materials-redo` branch. The member in question is in `HeatTransferBase`. Do we disable `-Wshadow`? Supposedly gcc 4.8 is better with this.

I tested this locally so I'm going to go ahead and merge right away.


roystgnr:
None of the default libMesh compiler flag settings enable -Wshadow, no.  Too many screaming warnings from inline functions in third-party headers, IIRC.

PR #310, opened 2015-09-20
More shadow fixes

pbauman:
Founds these as part of the work in `input-materials-redo` branch. I'd like to merge this ASAP, and I've tested, but I'd still feel better if @nicholasmalaya hit this. @nicholasmalaya are you using these Physics? If so, could you run a case make sure this doesn't change anything for you?

### DIFF
--- a/src/physics/include/grins/averaged_fan_adjoint_stab.h
+++ b/src/physics/include/grins/averaged_fan_adjoint_stab.h
@@ -65,10 +65,6 @@ namespace GRINS
 
   protected:
 
-    libMesh::Number _rho;
-
-    Viscosity _mu;
-
     IncompressibleNavierStokesStabilizationHelper _stab_helper;
 
   private:
--- a/src/physics/include/grins/averaged_turbine_adjoint_stab.h
+++ b/src/physics/include/grins/averaged_turbine_adjoint_stab.h
@@ -66,10 +66,6 @@ namespace GRINS
 
   protected:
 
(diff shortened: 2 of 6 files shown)


pbauman:
*ping* @nicholasmalaya

roystgnr:
Yikes!

This looks like an obvious atavism fix, and I can't come up with any way it would have led to a bug, though.  If we're not tripping the penalty poiseuille case then I think we should be safe to merge without double-checking old results.

nicholasmalaya:
I've been planning on looking at this but I won't be getting to it today. I'm actually not using these physics in production yet, but I could do some weaker tests. However, if it is not tripping @roystgnr regression for the fan, I think this should be good to go.

roystgnr:
You are using VelocityPenaltyAdjointStabilization in production, aren't you?

roystgnr:
I'm afraid all my fan + turbine regression tests are unstabilized.

nicholasmalaya:
> You are using VelocityPenaltyAdjointStabilization in production, aren't you?

Absolutely, however, when I looked at the changes I only saw edits to velocity_drag, not the velocity penalty proper. Is the drag function automatically being applied to that? I thought not.

roystgnr:
Ha!  I misread; you're correct.

roystgnr:
If we don't have *any* production runs with questionable code to double-check, then let's just merge the less-questionable code now.

PR #312, opened 2015-09-24
Parsed bcid fix

roystgnr:
This should fix some nasty ParsedBoundaryQoI bugs; previously adjoint and sensitivity solves wouldn't work and even QoI evaluation only worked correctly if your parsed function was zero on non-specified boundaries.

@vikramvgarg has some more tests to run before merge.


roystgnr:
Okay, this fixes some cases and doesn't seem to lead to new problems.

PR #275, opened 2015-04-07
Weighted flux qoi

roystgnr:
I want to add regression tests before merging, but the code should be in decent shape to examine.

PR #313, opened 2015-09-26
Add missing regression datafile for poisson_weighted_flux test

pbauman:
Looks like this was missing from #275. I went ahead and generated one from the current test.


roystgnr:
D'oh!

PR #314, opened 2015-09-28
Switch to ParameterMultiAccessor

roystgnr:
ParameterMultiPointer won't be flexible enough to handle the
non-pointer parameters in Antioch.

This is going in a separate PR so I can use the new API before the whole of #291 is done testing, and so avoid setting up future conflicts.

### DIFF
--- a/src/physics/include/grins/axisym_heat_transfer.h
+++ b/src/physics/include/grins/axisym_heat_transfer.h
@@ -87,7 +87,7 @@ namespace GRINS
     // class
     virtual void register_parameter
       ( const std::string & param_name,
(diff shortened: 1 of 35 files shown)


roystgnr:
Next time I "make check" *before* creating the PR.  Even without MooseBuild to yell at me it's still embarrassing.

pbauman:
Should we squash those together then for nicer history?

roystgnr:
Good point.

roystgnr:
Squished.

pbauman:
Cool. Merge when you're ready.

PR #315, opened 2015-09-29
Parsed function variable sensitivities

roystgnr:
This requires https://github.com/libMesh/libmesh/pull/667

We can now calculate sensitivities WRT inline variables in any parsed functions other than those imposed through boundary conditions.

@nicholasmalaya and @vikramvgarg might find this useful

This incorporates #314; we'll merge those two commits and pull that PR separately first and then rebase this off that; I just wanted to put the code out there for perusal now that it's passing tests with correct results.

### DIFF
--- a/src/physics/include/grins/averaged_fan_base.h
+++ b/src/physics/include/grins/averaged_fan_base.h
@@ -31,8 +31,8 @@
 #include "grins/inc_navier_stokes_base.h"
 
 // libMesh
-#include "libmesh/function_base.h"
 #include "libmesh/getpot.h"
+#include "libmesh/parsed_function.h"
 #include "libmesh/tensor_value.h"
 
 // C++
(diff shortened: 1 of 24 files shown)


roystgnr:
Rebased on squished #314

pbauman:
Oh wow, this looks rad. Once #314 gets merged to master, and this rebased, it ought to be easier to  read (still looks like there a bunch of MultiPointer/Accessor changes here).

roystgnr:
I'm still seeing the diffs from #314 even post-merge.  Not sure what to do to fix that.

roystgnr:
Okay, either the extra rebase on origin/master fixed things, or github finally caught up.  Either way, the massive changes in this PR are no longer obscured by the extensive changes from #314.

pbauman:
Good to go.

As an aside, I wonder if we need to tweak the API for init'ing Physics (change `auxiliary_init`?) since you're having to cache a GetPot object here in a couple of places. I feel like we shouldn't ever have to do that.

roystgnr:
We'd like two passes with GetPot available, I think: one to get the variables from each Physics so we can init the System, and a second so we can initialize objects that require a variable-populated System.

PR #316, opened 2015-10-02
Adjoint sensitivities include an adjoint solve

roystgnr:
So we shouldn't need to specify do_adjoint_solve manually in the input file before outputting z.


pbauman:
:+1:

PR #317, opened 2015-10-02
Refactor Input Specification for Material Properties

pbauman:
This supercedes #256.

This is not quite done yet, but 1. It's a monster and 2. There's a couple of points on which I'd like suggestions/feedback, so I'll go ahead and open now.

This is a monster because I wanted to do all the properties all at once so it was done and we could have consistency across the `Physics` and property classes.

The thread discussing the format of the input is [here](https://groups.google.com/forum/#!topic/grins-devel/AGVvODmfzh8). The gist is that all material properties now get tied to a material name. Then, within the `[Materials]` section, we can specify different properties for each material name. The name is set for each physics using `material_name`, except for `Physics` like stabilization classes, which get tied to their 'core' physics. This way, we can now specify different materials for different `Physics`.

In addition to the usual feedback, I'm soliciting feedback for the following:

1. @roystgnr, is it appropriate that the `AveragedFan` and the like are tied to `IncompressibleNavierStokes` or is it conceivable that we'd want to use those with other `Physics` (`Stokes`, `LowMachNavierStokes`,...)? If so, I'll need to check and update those to have them look for a `material_name`. The diff on `PhysicsFactory` will show you what I've tied to what (I added a `core_physics_name` option to the function call for each of the `add_physics` helper functions).

2. @nicholasmalaya, please hit this HARD. I've probably run `make check` over 300 times at this point, and the history is clean, but it's possible I've missed something. You should get a shit load of warnings, but all your results should be the same.

3. What do folks think is a good way to handle the reacting case? I made one suggestion in #256 where we still split the properties individually, but I was also thinking something like
<pre>
[Materials]
   [./5SpeciesAir]
      [./GasMixture]
         species = 'N2 O2 N O NO'
         thermochemistry_library = 'antioch' #could also be cantera
         [./Antioch]
             species_data = 'species.dat'
             kinetics_data = 'kinetics.dat'
             thermo_model = 'cea'
             thermo_data = 'thermo.dat'
             transport_model = 'mixture_averaged' # hopefully multicomponent someday
             transport_data = 'transport.dat' # for kinetics theory data
             viscosity_model = 'blottner' # or constant, sutherland, kinetics_theory
             viscosity_data = 'blottner.dat' #only needed for blottner, eucken
             viscosity_value = '1.0e-3' #only needed for constant
             thermal_conductivity_model = 'eucken' # or constant, constant_prandtl, kinetics_theory
             mass_diffusivity_model = 'constant_lewis' # or constant, kinetics_theory
         [../Cantera]
             cantera_data = 'cantera.xml' #holds kinetics, thermo, transport, and species data
             gas_mixture = 'air_5_species' #Cantera parses the XML looking for the mixture name
             transport_model = 'mix' #or god help you, multicomponent
</pre>
This and #256  are just two ideas. Feel free to chime in on what you think is the right way to do this. The main thing I don't like about splitting by property is that, right now, you're either all Cantera or all Antioch, no mixing and matching. It wouldn't be a *huge* effort to support that, but I also don't see the benefit either.

4. I'm still not happy with the actual input names being spread around, e.g. there's no central place where we say viscosity is `[Materials/MATERIAL_NAME/Viscosity/value]`. Any suggestions here? I was thinking we could park some more methods in `MaterialsParsing` where you pass the material name and get back the input variable, e.g.
<pre>
static std::string MaterialsParsing viscosity_input_str(const std::string& material)
{
   return std::string("Materials/"+material+"/Viscosity/value");
}
</pre>
so then it's in one place and we don't have it scattered around everywhere. Thoughts on this?
Things still to be done:
- [x] Update solid mechanics `Physics` to parse from `[Materials]`
- [x] Update any QoIs that parse material properties
- [x] Update example input files
- [x] Update thermochemistry properties once we settle on what it should look like
- [x] Finalize "special" flow classes (`AveragedFan` etc.) following feedback from @roystgnr 
- [x] Any fixes from @nicholasmalaya testing
- [x] If there's a consensus on centralizing the variable naming, implement it


nicholasmalaya:
Sounds good; I'll queue a few representative jobs and have feedback on Monday for you.

roystgnr:
AveragedFan et. al. *shouldn't* be tied to INS, but they *are* tied to INS, and it's not your job to fix that.  ;-)

pbauman:
> AveragedFan et. al. shouldn't be tied to INS, but they are tied to INS

OK, if we're respecting current assumptions, I'm OK with that.

pbauman:
To try and provide a little guidance on this:
- I created `PhysicsFactoryHelper` and `MaterialsParsing` classes that have only static members to house helper functions for parsing the material name, the model used for each property, etc.
- I updated one property at a time: viscosity, conductivity, specific heat, parameters in `LowMachNavierStokes`. Now moving onto solid mechanics.
- There may be some consolidation that I can still do for all the backward compatibility checking that's around. But that's the bulk of these changes --- doing all the checks for the options and erroring out if they collide, warn about using the deprecated options, and warn if falling back to a default value (I want to go to may past self and kick him in the nuts for ever allowing default values of material properties). 
- Other changes are just updating the input files in the testing to use the new method of parsing the material parameters. Note I always did that as a separate commit so as to test both the backward compatibility and the new capability.

nicholasmalaya:
Honestly this looks fine to me. I've run a series of old runs for several representative cases and everything looks good to me.  Timings are sane, results appear to be identical.

pbauman:
Great, thanks @nicholasmalaya

pbauman:
OK, this is finally ready to go. I've only touched reacting stuff since Nick tested (other than rebasing), so he should still be good.

Feel free to have one last look, otherwise I'll merge tomorrow afternoon if no one says anything. Note the final commit, 0e47815, I updated the "master" input file in an attempt to describe the usage.

If we come up with a better way to centralize the material parsing stuff, I'll be happy to implement it then, but since no one had an opinion, we'll go with what's here. 

FWIW, this is significantly more flexible and consistent than the insanity that was there before and is a much cleaner and clearer input file.

nicholasmalaya:
:+1:

roystgnr:
:+1:, thanks for the clarifications

pbauman:
Great, thanks all!

PR #318, opened 2015-10-02
Support Mesh/all_second_order after file read

roystgnr:
This lets us run first order meshes with higher order elements.

@vikramvgarg will be testing shortly


pbauman:
:+1: assuming tests go well

roystgnr:
Tests did not go well, but the flaw seems to be "numerical jacobians need more flexibility"; "mesh needs to be second order" worked perfectly.

PR #319, opened 2015-10-02
Add numerical_jacobian_h_variables/values options

roystgnr:
Requires https://github.com/libMesh/libmesh/pull/669 to compile.

Not currently exercised in regression tests, but works as expected on
my own input files

### DIFF
--- a/src/physics/include/grins/multiphysics_sys.h
+++ b/src/physics/include/grins/multiphysics_sys.h
@@ -181,6 +181,13 @@ namespace GRINS
     PhysicsList _physics_list;
 
     bool _use_numerical_jacobians_only;
+
+    // A list of names of variables who need their own numerical
+    // jacobian deltas
+    std::vector<std::string> _numerical_jacobian_h_variables;
+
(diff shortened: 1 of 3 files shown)


pbauman:
Cool.

Could maybe change one of the existing tests to set all the variables to have the same value as a not particularly strict test?

vikramvgarg:
This seems to be doing what its supposed to but not having any effect on the sensitivity results.

roystgnr:
Yup. Going from 1e-6 to 1e-9 on nu causes NewtonSolver convergence to stall out at ~300x residual reduction per step instead of ~3x, and it effects weighed flux sensitivities, but the effect seems to be that now the adjoint sensitivity is 55% too low instead of 50%.  Further shrinking nu_h helps nothing.  Tweaking h for the other variables helps nothing.

It works and it's useful for convergence, though, so let's keep it.  We recently added a "use ridiculously up to date libMesh" requirement for the parsed function sensitivities, so we might as well do so again.

PR #322, opened 2015-10-16
Allow WeightedFlux to use per-variable weights

roystgnr:
To make the syntax intuitive there must be one variable and one bc id
per weight, so now the syntax is more repetitive.

I haven't tested this much; @vikramvgarg make sure it doesn't regress your existing drag calculations before you try the angle-of-attack stuff.

### DIFF
--- a/src/qoi/include/grins/weighted_flux_qoi.h
+++ b/src/qoi/include/grins/weighted_flux_qoi.h
@@ -59,21 +59,6 @@ namespace GRINS
                        const MultiphysicsSystem& system,
                        unsigned int qoi_num );
 
-  protected:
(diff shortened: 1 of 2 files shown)


vikramvgarg:
Checked drag at AoA = 10 with NASA reference:

Cd_grins = 0.012660181
Cd_NASA_codes = 0.01225 - 0.01245

Should be good to merge.

vikramvgarg:
Oh also matches the existing 0 angle of attack result we had used as a regression earlier.

roystgnr:
Thanks!  :+1: when @pbauman is ready, then.

PR #323, opened 2015-10-26
Use ParsedFEM- not Parsed-Func in VelocityPenalty

roystgnr:
This may be necessary for some of @nicholasmalaya 's experiments.

This passes our VelocityPenalty regression tests but probably shouldn't be merged until Nick has time to play with it.

### DIFF
--- a/src/physics/include/grins/velocity_penalty_base.h
+++ b/src/physics/include/grins/velocity_penalty_base.h
@@ -31,7 +31,7 @@
 #include "grins/inc_navier_stokes_base.h"
 
 // libMesh
-#include "libmesh/parsed_function.h"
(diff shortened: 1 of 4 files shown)


nicholasmalaya:
Amazing, thank you @roystgnr. I'll get to testing this right away and report back.

nicholasmalaya:
@pbauman I'm recommending this pull be incorporated into grins master. 

It looks good in terms of regression, in that nothing broke. Furthermore, I've been using new functionality where by the penalty-value is velocity dependent (I've tested with 'u' and 'v'), and giving every indication that it is working as intended.

pbauman:
Cool with me, thanks.

PR #327, opened 2015-11-05
Fixes for libMesh/libmesh#701

pbauman:
cppclean was run on the libMesh tree and we needed to add a few
missing headers as a result.

### DIFF
--- a/src/solver/src/parameter_user.C
+++ b/src/solver/src/parameter_user.C
@@ -35,6 +35,7 @@
 #include "libmesh/parsed_function.h"
 #include "libmesh/parsed_function_parameter.h"
 
+
 namespace GRINS
 {
 #if LIBMESH_DIM == 3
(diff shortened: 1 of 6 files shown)


src/solver/src/parameter_user.C
```diff
@@ -35,6 +35,7 @@
 #include "libmesh/parsed_function.h"
 #include "libmesh/parsed_function_parameter.h"
 
+
```
roystgnr: Not quite enough white space?  ;-)

roystgnr:
:+1: ASAP; thanks!

src/solver/src/parameter_user.C
pbauman: HA! Totally missed that after I was screwing around to figure what was going on with the `ParsedFunction` business.

PR #329, opened 2015-11-20
Update to newest https://github.com/tsuna/boost.m4

roystgnr:
Using https://github.com/tsuna/boost.m4/commit/df28a91500c3263662868f3bc8912b2ed016972b fixes configuration for me on Ubuntu 15.10; with our old file configure dies after

```
checking for Boost headers version >= 1.39.0... yes
checking for Boost's header version... 
configure: error: invalid value: boost_major_version=
```

### DIFF
--- a/m4/common/boost.m4
+++ b/m4/common/boost.m4
@@ -1,5 +1,5 @@
 # boost.m4: Locate Boost headers and libraries for autoconf-based projects.
-# Copyright (C) 2007, 2008, 2009, 2010, 2011  Benoit Sigoure <tsuna@lrde.epita.fr>
+# Copyright (C) 2007-2011, 2014  Benoit Sigoure <tsuna@lrde.epita.fr>
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


m4/common/boost.m4
```diff
@@ -107,13 +126,6 @@ AC_ARG_WITH([boost],
    [AS_HELP_STRING([--with-boost=DIR],
                    [prefix of Boost $1 @<:@guess@:>@])])dnl
 AC_ARG_VAR([BOOST_ROOT],[Location of Boost installation])dnl
-# If BOOST_DIR is set and BOOST_ROOT is not, use BOOST_DIR.
```
pbauman: I think we'd added this back in the day. Do we want to keep it around?

roystgnr: Thanks, I missed that!

I'd actually prefer to remove it, though.  BOOST_ROOT seems to be "standard" enough, and I'm in the middle of remaking modules so it's easy enough to define that in addition to BOOST_DIR.

Your preference?

pbauman: Agreed, let's leave it out.

pbauman:
I'll give this a whirl on my laptop this afternoon and then will merge.

pbauman:
Cool. Need to update the config_summary, but I can do that separately. Thanks!

PR #330, opened 2015-11-26
Make Boost Optional

pbauman:
I've wanted to do this for awhile, but the straw that broke the camel's back was needing to use GCC 5.2.0 on my laptop (thanks for breaking backward compatibility with the assembler Apple) and there were *many* warnings about the `auto_ptr` interface in the Boost smart pointer implementations, making real warnings undetectable.

This PR adds `GRINS::SharedPtr`. I took a cue from the discussion libqueso/queso#398 and went with inheritance instead of the template typedef trick. We piggy back on libMesh's check for `std::shared_ptr` and prefer that if it's there. If not, we fall back to Boost, and then error if `boost::shared_ptr` is not available. I also updated all uses of `boost::scoped_ptr` to use `libMesh::UniquePtr`. Note I haven't touched the `libMesh::AutoPtr` uses, we'll do that in a separate PR. Finally, updated the build system so we only try to detect (and use) Boost if libMesh didn't find `std::shared_ptr`. I also updated the config_summary  for the Boost changes. Also updated config_summary to make the info reported on Cantera and Antioch a little more precise/verbose while I was at it.

Happy Thanksgiving GRINS users and developers.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -136,7 +136,7 @@ dnl----------------
 dnl Check for boost
 dnl----------------
 BOOST_REQUIRE([1.39.0])
-
+BOOST_FIND_HEADER([boost/shared_ptr.hpp])
 
 dnl-------------------------------------
 dnl Check for GNU Linear Programming kit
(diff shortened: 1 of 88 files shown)


roystgnr:
That was all much more straightforward than "88 files changed" lead me to fear.  :+1:

pbauman:
OK, I dropped the unnecessary destructors and squashed that commit. Thanks!

PR #332, opened 2016-01-04
Consolidate Variable Initialization/Usage

pbauman:
This effectively completes what was started in #73, namely encapsulating the variable name parsing and variable initialization with the `libMesh::System` into standalone classes. Then, updating `Physics` classes to only use these objects to handle variables. This is opposed to the scattered mess that was there before. The design follows as before: `*Variable` classes for basic handling of variable name and index (primarily used in `BCHandling` classes) and `*FEVariable` subclasses for the finite element space (used in the `Physics` classes).

The first set of commits updates the `Physics` and `BCHandling` classes that could've used already existing `(FE)Variable` classes. Then, I added thermodynamic pressure and species mass fraction versions of the `(FE)Variable` classes and then used them. The vast majority of the addition/deletions in this PR are search replace from hardcoded variable index to using the `Variable` object.

I believe I got everything except the QoIs (and there's only a couple that need changing) and two specialized `Physics` classes. The QoIs currently don't take a GetPot object in the constructor, so I can drop these objects in there yet. It seemed overkill to create such `Variable` classes for the specialized `Physics`.

This is the precursor to redoing the input file sectioning for the FEM variables, ala the discussion in [this grins-devel](https://groups.google.com/forum/#!topic/grins-devel/AGVvODmfzh8) thread. After this, we should be able to limit the changes needed to (slightly more than) a handful of classes.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -139,6 +139,7 @@ libgrins_la_SOURCES += physics/src/parsed_source_term.C
 libgrins_la_SOURCES += physics/src/thermo_pressure_variable.C
 libgrins_la_SOURCES += physics/src/thermo_pressure_fe_variable.C
 libgrins_la_SOURCES += physics/src/species_mass_fracs_variables.C
+libgrins_la_SOURCES += physics/src/species_mass_fracs_fe_variables.C
 
 # src/properties files
 libgrins_la_SOURCES += properties/src/parsed_property_base.C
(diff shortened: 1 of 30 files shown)


roystgnr:
:+1:

Only qualm: reading it over and over again makes the last _var suffix in ```_foo_vars.f_var()``` feel more and more redundant.

pbauman:
Thanks. I'm happy to change the `_foo_vars.f_var()` to something else if you have some suggestions.

roystgnr:
How about _foo_vars.f()?

pbauman:
Will do. OK to merge after that set of changes?

pbauman:
Actually, I just realized that change would be a nasty rebase for #333. Mind if I make the `_var()` removal changes there?

roystgnr:
Sounds fine, yeah.

pbauman:
Great, thanks.

PR #333, opened 2016-01-05
Variable Class Refactor

pbauman:
Builds on #332, will rebase once that's merged.

I started redoing the input file spec for the `System` variables and decided to scratch this itch. 

First was just aesthetic changes: putting the `Variables` classes in their own directory and renaming `SolidMechanics` to `Displacement`. Then, introduce lightweight subclasses to use a single common container for all the variable quantities. This will save some lines of code in the input refactoring, give a place to put common deprecated functions, and is probably the better way to do this (even if maybe over-engineered a bit).


roystgnr:
Going to be lazy and wait to look at diffs until after #332 is merged.

pbauman:
Rebased. Will plop the `_var()` removal on here as discussed in #332.

pbauman:
Done. I <3 sed. I can squash them all together if there's a preference for that. (I was leaning that way, but figured I wait and see what others thought first.)

pbauman:
I'd like to merge this by Jan. 17, if folks can find time by then, in order to move forward with the refactoring of the input parsing for the variables (which I ought to be able to do starting the 17th). I don't want to proceed there until folks chime in here one way or other.

pbauman:
Going once...

pbauman:
Going twice...

roystgnr:
Sorry about the delay; looking it over now.

pbauman:
> Sorry about the delay; looking it over now.

No worries, thanks.

roystgnr:
:+1: as is, but unless I'm mistaken, you could shave off a bunch of code still.

pbauman:
> Oh, wait, that's what you did. So are e.g. _TU_order and _TU_FE_family redundant with these?

Right, if there are any of those left, they would be redundant. (This was my comment about maybe slightly overengineering this, but it will make variable-input-refactoring cleaner). But your comment 

>would it be possible to move init() into the base class? Loop over _var_names, and for each name assert(has_variable) then set _vars.

was spot on, I think. We ought to be able to do that. I'll go back and clean that up. Any other obvious things for shrinking?

pbauman:
Also, I was thinking squashing 59fdf16 though d895bef made sense. Do you agree?

roystgnr:
Squashing isn't necessary but definitely makes sense.

pbauman:
OK, I rebased on master, squashed the aforementioned commits, and added a common method to each of `VariablesBase` and `FEVariablesBase` to be used by subclasses that should've shaved a few dozen lines (I had to add new .C files for the implementation of those methods, so there were raw source lines added due to source headers). I went this route instead of full inheritance since we would've ended up in multiple inheritance hell with the `init()` method or adding a few shim classes to avoid that (and not very cleanly), so this seemed like the best compromise.

Will merge tonight if I hear no complaints.

PR #336, opened 2016-01-15
__LIBMESH_<DATE,TIME>__ are no longer defined

pbauman:
And that's OK because we were *lying*. The source was built at __DATE__
and __TIME__ since this is GRINS and not libMesh.


roystgnr:
Ha!  I didn't even think about that aspect of it.  :+1:

PR #337, opened 2016-01-21
Update Examples to use parsed_dirichlet Boundary Condition

pbauman:
Instead of outdated `ParabolicProfile`


pbauman:
Tested that both examples work with changes, so merging right away.

PR #338, opened 2016-01-21
Fix test that was failing with PETSc >= 3.6.x

pbauman:
The change was in the default behavior on the shift type for PC ilu.
See this thread: http://lists.mcs.anl.gov/pipermail/petsc-users/2016-January/028210.html

@dknez I have a vague memory of you mentioning some failures that happened with PETSc >= 3.6.0, but I can't find the thread now. I'd bet this was the culprit. If I'm misremembering, sorry for the noise.

Merging right away.


dknez:
I don't remember the thread about 3.6.0 failures, so I'm not sure if that was from me or someone else.

PR #339, opened 2016-01-23
Refactor unsteady solver options

pbauman:
The is part 1 of the refactoring of parsing solver options. Overall, the interim plan is to separate the "solver" (e.g. linear solver, nonlinear solver, time stepping, etc.) options from "strategy" options (AMR, adaptive time stepping, continuation, etc.). To that end new classes (mostly with static methods) have been created for parsing those various things.  This PR is only focused on the time stepping options. Thus, the `StrategiesParsing` is very trim right now and, in particular, I haven't updated those options yet.

A nice side-effect is that we can now specify the time solver instead of only using `EulerSolver`. Now you can use `Euler2Solver`! :P But this should have a lot of the groundwork for easily extending this.

So, the new input options for time stepping look like:
<pre>
[SolverOptions]
   [./TimeStepping]
      solver_type = 'libmesh_euler_solver'
      delta_t = '0.1'
      n_timesteps = '10'
      theta = '1.0'
[]
</pre>

There are no defaults for `delta_t` and `n_timesteps` and, thus, you get an error if you don't have them. To trigger an unsteady solve, we look for `[SolverOptions/TimeStepping/solver_type]` --- if it's not set, then an unsteady solver will not be used. The user would get UFO errors then about other `TimeStepping` options in that case (unless they foolishly disabled that...). The default for `[SolverOptions/TimeStepping/theta]` is `0.5`. 

All the tests pass with and without these changes through the whole history of the PR. Thus, I feel pretty good about this being backward compatible, particularly since there's several tests hit by these options.

### DIFF
--- a/examples/master_example_inputfile.in
+++ b/examples/master_example_inputfile.in
@@ -326,3 +326,42 @@
             # Poisson's ratio
             nu = '0.3'
 []
+
+
+# The block below illustrates specifying options for the solver.
+# Solvers are things like linear solver, nonlinear solver, and
+# time stepping.
+[SolverOptions]
+
+   # Options related for time stepping algorithms
+   [./TimeStepping]
+
+      # This is the algorithm for time-stepping. Time stepping
+      # is triggered by setting this option, so if it is not
+      # set, then an unsteady algorithm will not be selected.
+      #
+      # Current valid options are:
+      #   libmesh_euler_solver
+      #   libmesh_euler2_solver
+      solver_type = 'libmesh_euler_solver'
+
+      # This sets the time step size. There is no default.
+      # If using an unsteady algorithm, this option must be
+      # set or an error occurs.
+      delta_t = '1.0'
+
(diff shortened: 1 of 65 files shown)


pbauman:
Just FYI, the vast majority of the 65 files changed was the input files for the regression tests.

pbauman:
@nicholasmalaya, gives this a whirl when you have a chance, please?

pbauman:
I'd like to merge this ASAP. I feel pretty good about the coverage of the solver options change that I don't think it's critical for @nicholasmalaya to hit (`make check` passes through the whole history since I developed it that way). If y'all don't like the new option names or you don't like how I split out the parsing, speak by tomorrow night or forever (or until we redo it all again...) hold your peace.

nicholasmalaya:
I say go for it, I'm too busy as it is so I won't get to it until the weekend, and it sounds like it's likely safe.

pbauman:
Rebased.

PR #340, opened 2016-01-29
Physics Names Refactor

pbauman:
This PR does 3 things:

1. Move `grins_physics_names.h` to `physics_naming.h`
2. Get rid of `PhysicsName` typedef that was really stupid and noobish (one of my first of many regrettable choices)
3. Put all the physics names inside a new `PhysicsNaming` class and use static methods to access the string.

This is much cleaner, I think (we can easily see the physics names in Doxygen now!), but also sets up the multiply-enabled-physics refactoring to not have to change much code. 9affd04 is a monster, but was generated through judicious use of commands like

<pre>
for i in heat_transfer_source heat_transfer heat_conduction scalar_ode spalart_allmaras_spgsm_stab spalart_allmaras averaged_turbine averaged_fan_adjoint_stab averaged_fan parsed_velocity_source_adjoint_stab parsed_velocity_source velocity_penalty3_adjoint_stab velocity_penalty2_adjoint_stab velocity_penalty_adjoint_stab velocity_penalty3 velocity_penalty2 velocity_penalty velocity_drag_adjoint_stab velocity_drag incompressible_navier_stokes_spgsm_stab incompressible_navier_stokes_adjoint_stab incompressible_navier_stokes stokes; do srcgrep $i | grep -v PhysicsNaming | grep -v "#include" | grep -v static | cut -d : -f 1 | xargs sed -i "s/GRINS::$i/GRINS::PhysicsNaming::$i()/";  done
</pre>

where you only have to change the `sed` arguments for each of the patterns that are used around the physics name. (Sadly, I didn't figure that out until about 1/3 of the way in. Then it only took 10 mins. to finish and have `make check` passing.)

While this will break any derived applications, it does *not* change any runtime usage. I think I'm the only one that has any derived apps, and they're fixed with a handful of `sed` commands. Thus, I feel this is safe to merge, barring any disagreements about the new refactoring.


roystgnr:
I didn't think the typedef was noobish at all, actually.  Overengineered, but that's a good thing.  The PhysicsNaming bit is a good idea, though.

pbauman:
You want to keep to keep `PhysicsName` typedef, then? Easy enough to drop 3a612fa from the PR.

roystgnr:
Sure.  Feels more self-documenting to me.

pbauman:
Done. Thanks!

PR #342, opened 2016-01-29
Multiply Enabled Physics From Input

pbauman:
This PR enables using `:` in the input physics name and putting whatever suffix you want. This will allow us to use the same physics in different ways. For example:

<pre>
[Physics]
   enabled_physics = 'ElasticMembrane ElasticCable:Stringer ElasticCable:RingCable]
   [./ElasticMembrane]
   [../ElasticCable:Stringer]
      enabled_subdomains = '0'
      material = 'Kevlar'
   [../ElasticCable:RingCable]
      enabled_subdomains = '1'
      material = 'ThickKevlar'
[]
</pre>

The way I did this is by stashing the suffix in `PhysicsNaming`. The returned physics name will always have a suffix appended to it; the suffix is empty until the user sets it. For each physics we build, we first extract the suffix, set it, then it will propagate all the way through during the build phase of the physics, e.g. for stabilization classes that are tied to a "core" physics and its options. Then, when the building of the physics is done, we clear the suffix.

This is not 100% ready yet, hence marked do not merge. I need to refactor the variable naming first before merging this. Since variables are still tied to the physics section, you get surprises if you don't have matched suffixes between the physics. (The variable input refactoring shouldn't take long.) In the meantime, this can at least have eyes on the principal implementation (and the principle of the implementation). Then, once the variables bit is done, I'll throw in a few more tests here (and fix any lingering problems that might crop up). Then should be good to go.

Since we only do parsing in the physics at construction time, I think we still satisfy the principle of least surprise (modulo the variables input refactoring), but it's certainly possible I overlooked something.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -129,6 +129,7 @@ libgrins_la_SOURCES += physics/src/nonlinear_elasticity_instantiate.C
 libgrins_la_SOURCES += physics/src/source_term_base.C
 libgrins_la_SOURCES += physics/src/constant_source_term.C
 libgrins_la_SOURCES += physics/src/parsed_source_term.C
+libgrins_la_SOURCES += physics/src/physics_naming.C
 
 
 # src/properties files
--- a/src/physics/include/grins/physics_factory.h
+++ b/src/physics/include/grins/physics_factory.h
@@ -78,6 +78,18 @@ namespace GRINS
     virtual void check_physics_consistency( const GRINS::PhysicsList& physics_list );
 
     //! Utility function
+    /*! The elements in the physics_list may have suffixes attached while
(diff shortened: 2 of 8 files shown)


pbauman:
Awesome. I created a branch that rebased this on #345. Works perfectly --- I can set different, non-slave, `Physics` with different delimiters and everything Just Works. Bam. As soon as #345 is merged, I'll rebase this and cherry-pick the other tests onto this and merge.

pbauman:
OK, rebased and cherry picked a couple more input updates to test the suffix treatment. This is good to go as far as I'm concerned. I'll let this sit a bit and work on some other branches, but I would like to merge this quickly to do some refactoring on the test structuring.

PR #345, opened 2016-02-02
Variable Input Refactor

pbauman:
This PR contains the refactoring of the parsing for the variable names and FE family/order. Now, instead of being tied to the `Physics`, the variables are in a standalone section. This makes much more sense because there may be multiple `Physics` that reuse the same set of variables. There was a fairly substantial internal refactoring here (again) to take advantage of 1. single variables (e.g. temperature) and 2. single FE type (all FE type/order are the same for all variables). This made all the leaf classes much trimmer and I think was a big improvement.

To get a feel for the new format, here's an example from one of the reacting flow inputs:

<pre>
[Variables]
   [./Velocity]
      names = 'u v'
      fe_family = 'LAGRANGE'
      order = 'SECOND'
   [../Pressure]
      names = 'p'
      fe_family = 'LAGRANGE'
      order = 'FIRST'
   [../Temperature]
      names = 'T'
      fe_family = 'LAGRANGE'
      order = 'SECOND'
   [../SpeciesMassFractions]
      names = 'Y_' #Here, names is the prefix to each species
      fe_family = 'LAGRANGE'
      order = 'SECOND'
[]
</pre>

In a subsequent PR (with `ConvectionDiffusion`), I'll add a "GenericVariable" that will be tied to the physics name; this is where we can park things like `average_fan`, etc.

I feel pretty good about the coverage here, so @nicholasmalaya don't feel pressured to run this through your gauntlet (but feel free to if you have time). I'll let this percolate for a bit as I'll rebase #342 on this and make sure that's working with these changes as I thought it should.


pbauman:
#342 works well with this. Going to merge this this afternoon unless others complain.

PR #348, opened 2016-02-04
Add Knob for System::extra_quadrature_order

pbauman:
@vikramvgarg, please confirm this works for you. You'll need to add the following to your input file:

<pre>
[Strategies]
   [./Assembly]
      # Value must be >= 0
      extra_quadrature_order = 1 
[]
</pre>

If you have a self-contained input file that I could use to generate a regression test, that would also be nice.


vikramvgarg:
Yes, that fixed the problem. The adjoint error estimates now match for GRINS and libMesh. 
[quadrature_test.txt](https://github.com/grinsfem/grins/files/116915/quadrature_test.txt)

Have attached the test I did as .txt file, github does not like .in it seems ...

pbauman:
Holy cow! 15 extra orders!

Thanks for the test. I'll add it onto this PR in the morning and then merge.

pbauman:
Test added. For this test to pass, the local libMesh installation needs to be at least at hash libMesh/libmesh@311deb16f46aa2f5cf68bf54a2f32944430d450c

PR #355, opened 2016-02-23
Test Restructuring

pbauman:
There are three primary components to this PR. The first is removing as much of the configure-time setup of the testing as possible (e.g. the `AC_CONFIG_FILES` call on the input files and shell scripts). Now, you can just use the environment variables defined in `AM_TESTS_ENVIRONMENT`in your .sh file and no need for touching configure.ac. We had to symlink the grid files and a few others so the tests could access them at check time.

The second chunk of this PR is reorganizing the tests for some future work. Now, inside the `test` directory, there is `error_xfail` (tests that we expect to fail because we're checking that we're erroring where we want to), `exact_soln` (tests for PDEs where we have an exact solution), `interface` (tests for interfaces to external libraries, e.g. Antioch, Cantera), `regression` (cases where we have "gold" data only, no exact solution), and `unit` (true blue "unit" tests). Right now, the tests have only been moved around. 

The final chunk of the PR is getting `make distcheck` working with all these changes. Because of the way `AC_CONFIG_LINKS` works, this turned out to be a bit of a PITA.

In subsequent PRs, I want to add CppUnit and move the `unit` and `interface` tests to use CppUnit. Additionally, I'll add a new test for the exact solution were we run the `grins` app directly and then check against the exact solution (parsed from the command line). Finally, and again subsequent to this, clean up all the copy-pasta crap in `regression` and run the `grins` app directly and check against the gold data.


pbauman:
Discussed with @roystgnr in meat space and good to go here.

PR #356, opened 2016-02-24
Generic Exact Solution Test

pbauman:
Builds on #355, will rebase once that's merged.

This add a "generic" test whereby you first run your input file through GRINS, writing out a restart file of the solution. Then, you run this new test which will compare your solution file with the exact solution (parsed using `ParsedFunction`). You prescribe the expected error and the tolerance for the mismatch between your specified error and the computed error. That is, if the error is not zero, say 1.234e-9, then the user would supply 1.234e-9 as the expected error and then the tolerance is the allowed deviation from that value.

Redid all but one of the `exact_soln` tests using this.

Also, first PR from me at 38000 ft.


pbauman:
Rebased on master following #355 merge. Should be easier to look at this now.

pbauman:
Added one more comment to README about the time-dependent case.

pbauman:
The only remaining question on this PR in my mind is how to handle QoIs. I was trying to avoid using the `GRINS::Simulation` in the test, but maybe it's unavoidable.

pbauman:
Talk to @roystgnr in meat space. Consensus was to enable writing the QoI to a text file and reading it. Will do that in a separate PR.

PR #358, opened 2016-02-24
Reinit constraints for time dependent Dirichlet boundary conditions

pbauman:
Plus an `exact_soln` test that exercises it. Builds on #357, will rebase once that's merged.

PR #359, opened 2016-02-28
Fix for time-dependent Dirichlet boundary conditions

pbauman:
Replaces #358. History changed and it was easier just to cherry-pick.

We were only calling reinit_constraints for nonlinear Dirichlet boundary conditions, but not for time-dependent ones. This fixes that and adds an `exact_soln` test that exercises the changes (based of a libMesh example).


roystgnr:
This doesn't include the workaround that Junchao Zhang came up with (to fix the problem where u_old gets adjusted using the constraints at t_new) on the mailing list, does it?  It would be interesting to see how big an effect that has on the comparison to exact solution.

pbauman:
Correct, it does not. My goal was to get a test in place so we could see exactly these differences. That's the next PR (hopefully today) since we need transient AMR to trigger.

roystgnr:
No transient AMR required!  If I understand the bug correctly, all that's required to trigger it is the use of true DirichletBoundary with a time-dependent function.

PR #363, opened 2016-03-07
Add CppUnit Optional Testing Dependency, Update Thermo Interface Tests to Use CppUnit

pbauman:
The first bit of this PR adds a check for CppUnit and adds paths accordingly to the test/Makefile.am. The only catch here is: 1. If your system is GCC 4.x based and you're using GCC 5.y, you need to build your own CppUnit (because of the cxx11 business in the STL headers that started in GCC 5.y)  and 2. If you're on a GCC 5.x based system, you'll need to build your own CppUnit if you want to use Clang (because of current inconsistencies in libstdc++ and libc++). I've already gotten used to having a cppunit module tied to the compiler because of the first exception, so it's not that big of deal. But comments welcome @roystgnr if you see a better way of handling that.

The second part of the PR introduces CppUnit based testing for the thermo interfaces in Cantera and Antioch. Currently only testing NASA9. But gives a good example of how to use this and so I wanted to get this merged. More testing to come (kinetic rates next!).

Note that this builds on #362.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -306,6 +306,17 @@ AM_CONDITIONAL(CANTERA_ENABLED,test x$HAVE_CANTERA = x1)
 
 AX_PATH_ANTIOCH(0.4.0,no)
 
+# -------------------------------------------------------------
+# cppunit C++ unit testing -- enabled by default
+# -------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


pbauman:
Rebased and merging.

PR #364, opened 2016-03-09
Kinetics CppUnit-Base Testing

pbauman:
Builds on #363.

Replaces a bunch of old crappy regression tests with true unit tests. I constructed these air tests by hand, working my way towards the automated code that's here in about 3 stages. We're within reasonable epsilon factors for Antioch. Cantera required a much more lax tolerance - not sure what the discrepancy is.

I've tried to make the testing functions as reusable as possible. Namely, drop in whatever species data you want (thinking NASA thermo data) in the base classes (`SpeciesTestBase`, `NASA7ThermoTestBase`, `NASA9ThermoTestBase`). Then, the child classes will designate what species are active in the test, the reaction data, etc. So, in theory, I ought to be able to drop in an ozone.C, similar in spirit to the air5sp.C in this PR, and reuse all the underlying functionality for the test. (In fact, I plan on doing the ozone one next as there's been some very strange discrepancies there, but may need to add some more Antioch parsing functionality first.)

Note I did not hold back on class hierarchies here (it is a C++ unit testing framework after all), but I've tucked everything in the `GRINSTesting` namespace, so it's easy to pick out in the Doxygen documentation; there is a commit here that makes sure that documentation always gets built. Nevertheless, I've tried to make the computational parts look very close to the math, performance be damned (but not too much, test still runs very quickly).


pbauman:
Rebased. Gonna sit on this one until @tradowsk and I sort out some weirdness with an Ozone test that builds directly on this.

pbauman:
OK, @tradowsk found out the issue (my bug): wasn't properly initializing indices in the test. Will add a fixup commit with those and this will be ready to merge.

@tradowsk, branch from the tip of this branch for your Ozone updates and we'll rebase when it's ready to go.

pbauman:
Fix for species index initialization committed and squashed. This is ready to go once we merge #368.

pbauman:
Rebased on master after #368 merged. This is good to go. Will merge at lunch unless comments to the contrary.

PR #367, opened 2016-03-15
adding scalar parsed qoi test case

nicholasmalaya:
Talked with @roystgnr about this briefly, and we both agreed this should be a tested capability. 

This case is almost a complete duplicate of the parsed_qoi.in test, but has been modified to ensure the parsed_qoi parsing appropriately handles scalars over the interior of the domain. This integral is analytically 5.0, and my machine is reporting,
<pre>
==========================================================
parsed_interior = 5.0000000000000648e+00
==========================================================
</pre>

e.g. 6.48e-14 error. Appears quite sane. However, any easy way to add this fact to the regression shell script? Right now it is just running, not testing this equality.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 Version 0.7.0 (In progress)
    * Added support for output of 'base_velocity' in averaged_fan
+   * Added scalar parsed qoi regression test case
 
 Version 0.6.1
 https://github.com/grinsfem/grins/releases/tag/v0.6.1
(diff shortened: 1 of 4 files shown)


pbauman:
Thanks for this. I need to add functionality to handle the QoI testing bit (briefly mentioned in #356). I'll make this a priority.

roystgnr:
Let's merge this for now, and update it with the others when we add QoI regression testing?

pbauman:
Sounds good. Sorry I haven't been able to get the QoI testing in yet.

PR #372, opened 2016-03-19
String utils testing

pbauman:
Adds a handy string utility function and CppUnit testing for the tests in the unit directory and CppUnit-based tests for some of the string utility functions. Have not converted other unit tests to CppUnit at this point.


pbauman:
Merging right away as this is a simple thing and `make check` passes.

PR #373, opened 2016-03-19
CppUnit Driver Update to include libMesh

pbauman:
Updated the CppUnit driver in `test/unit` to have an MPI communicator so we can start doing more interesting tests. I have in mind using this downstream in some variable refactoring and boundary condition parsing refactoring, but this at least sets up the basic infrastructure and adds an example test for `MeshBuilder`.


pbauman:
`make check` is passing with this. Merging right away.

PR #374, opened 2016-03-19
Variable Init Refactor

pbauman:
Refactored how we handle init'ing the `VariableBase` vs. the `FEVariablesBase`. What was there before was confusing. Now, we have an `init_vars` (extracting variable numbers from the `System`) for `VariablesBase` and `init` (adding variables to the `System`) for `FEVariablesBase`. 

Added CppUnit-based testing for most of the variable classes. That addition included adding a testing helper class for setting up a basic `MultiphysicsSystem` for these kinds of tests.

This refactoring is to promote the use of `VariableBase` factories downstream, which will then in turn be used in boundary condition registration refactoring.


pbauman:
All the tests pass, so merging right away.

PR #375, opened 2016-03-19
Factory Restructure

pbauman:
Add a couple of low-level abstract factories so that we can reuse things like `set_getpot` across many different factories.


pbauman:
Tests passing. Merging.

PR #376, opened 2016-03-19
Variable Factories

pbauman:
This PR adds factories for building `VariableBase` objects. We don't use these in the `Physics` as that's not the intended use case. Instead, we will use these to extract variable names during the boundary condition construction (downstream of this PR). This included CppUnit-based testing of several of the factories and testing of the use of `init_vars` in the intended use case.


pbauman:
Merging this right away as this feature includes unit tests that pass.

PR #378, opened 2016-03-22
Constrained Variables and Variable Registration

pbauman:
This PR adds two things:

1. The notion of a "constrained" variable to `FEVariableBase`
2. `VariableWarehouse`

Both are unit tested in the CppUnit-based test. Both changes are motivated by downstream chagnes in the boundary condition refactoring. 

The first change is to help with determining if a `Variable` requires a boundary condition specified or not. (The plan downstream is, in the new format, every boundary must have a boundary condition for every `Variable`, even if it's just `natural`, to help reduce input file bugs with the boundary conditions.) Constrained variables, e.g. Lagrange multiplier type things, do not require boundary conditions, so the first change allows you specify at construction time whether or not the variable is a constrained variable. Then, we'll be able to determine later if it needs a boundary condition.

The second change is also motivated by downstream changes in the boundary conditions. Originally, I was using the `VariableFactory` to build `VariableBase` objects, but `VariableBase` objects don't have enough information, e.g. the constraint variable bits I just added or, in the case of species mass fractions, I really need to know the material in order to be able to pull out the chemistry library being used for constructing things like catalytic walls. So, the `VariableWarehouse` gives us place to fetch const information about each of the variables registered. We use the centralized naming in `VariablesParsing` so everything should be consistent. I've tucked `VariableWarehouse` in a `GRINSPrivate` namespace because right now, I do not allow duplicate variables to be registered. Once we support per-subdomain variables, that may change and we may need to update the API. So I put this in GRINSPrivate to indicate that the API could be in flux.

As as aside, once the boundary condition refactoring is done (today, hopefully), and along with this PR, we can get rid of `VariableBase` altogether since that was originally needed to provide info to `BCHandling` (which will die with the boundary condition refactoring). The `VariableWarehouse` will give us access to `FEVariableBase`, so we'll be able to reduce the code in the `Variable` design and have `FEVariableBase` be the root class (which makes much more sense I think).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -219,6 +219,7 @@ libgrins_la_SOURCES += variables/src/species_mass_fracs_variables.C
 libgrins_la_SOURCES += variables/src/variables_factory_base.C
 libgrins_la_SOURCES += variables/src/variables_factory.C
 libgrins_la_SOURCES += variables/src/variables_factory_with_physics_name.C
+libgrins_la_SOURCES += variables/src/variable_warehouse.C
 
 # src/utilities files
 libgrins_la_SOURCES += utilities/src/grins_version.C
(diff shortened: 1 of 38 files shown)


pbauman:
OK, I need this now for the boundary conditions, so I'm going ahead and merging. Apologies @roystgnr this didn't get to sit for the full 2 days.

PR #381, opened 2016-03-24
Move DisplacementContinuationSolver to Examples

pbauman:
This was kind of specialized anyway and will serve as an example for subclassing the solver to do fun things.


pbauman:
Going ahead and merging this. This is the last PR before the workshop. :( Workshop materials will be based on the tip of master after this PR.

PR #387, opened 2016-04-12
Add missing init_context() overrides

roystgnr:
This should help prevent possible bugs if users choose stranger combinations of physics operators.

I was working on this preemptively to also fix any bugs triggered by https://github.com/libMesh/libmesh/pull/892 - however if there are any such, we don't hit them in the test suite.


pbauman:
Thanks!

PR #389, opened 2016-04-13
fix for static initialization issues in physics_naming

tradowsk:
`make check` failed all test due to segfaults with GCC 4.8.4. @pbauman traced issue back to static initialization fiasco in Physics Naming.

Encapsulated static `_suffix` variable in static function to fix.

All tests now passing.

### DIFF
--- a/src/physics/include/grins/physics_naming.h
+++ b/src/physics/include/grins/physics_naming.h
@@ -36,11 +36,11 @@ namespace GRINS
   {
   public:
 
-    static void set_suffix( const std::string& suffix )
-    { _suffix = suffix; }
+    static void set_suffix( const std::string& suff )
+    { PhysicsNaming::suffix() = suff; }
 
     static void clear_suffix()
-    { _suffix.clear(); }
+    { PhysicsNaming::suffix().clear(); }
 
     //! Extract the physics name and the suffix from the full_name
     /*! Note returned suffix includes the delimiter */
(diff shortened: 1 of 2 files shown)


src/physics/include/grins/physics_naming.h
```diff
@@ -60,140 +60,144 @@ namespace GRINS
...
-    { return "ConvectionDiffusion"+_suffix; }
+    { return "ConvectionDiffusion"+_suffix(); }
 
   private:
 
     static std::string physics_name_delimiter()
     { return ":"; }
 
-    static std::string _suffix;
+    static std::string& _suffix()
```
pbauman: Please name the function `suffix`.

src/physics/include/grins/physics_naming.h
```diff
@@ -60,140 +60,144 @@ namespace GRINS
...
+    {
+      static std::string suffix;
```
pbauman: Please name this `_suffix`.

src/physics/include/grins/physics_naming.h
```diff
@@ -37,10 +37,10 @@ namespace GRINS
   public:
 
     static void set_suffix( const std::string& suffix )
-    { _suffix = suffix; }
+    { _suffix() = suffix; }
```
pbauman: Let's be explicit: `PhysicsNaming::suffix()` (assuming the function name change).

pbauman:
Thanks! Please squash in those changes, post here that you pushed (so I get the email) and then it should be good to merge.

Note I also tripped this when I tried tip of master with GCC 4.8.5. (I've been developing with GCC 5.3.0.). Looks like we'll need a few compiler versions in the CI (server is supposed to arrive at UB on 4/22).

src/physics/include/grins/physics_naming.h
tradowsk: If I change the function name to `suffix`, then it clashes with the parameter`suffix` on line 39 (`set_suffix()`). Any preferences for what to rename that?

pbauman: > If I change the function name to suffix, then it clashes with the parametersuffix on line 39 

Not if you call the function as `PhysicsNaming::suffix()`

tradowsk: Yeah, this note wasn't up yet when I saw your other comment. Apologies

tradowsk:
Made the requested changes, and verified `make check` still passing

pbauman:
Congrats @tradowsk on your first `fixup` commit and thanks for the fix.

PR #390, opened 2016-04-14
Fan velocity at a point is proportional to r

roystgnr:
These were supposed to be examples of rigid fans, not funky whirlpool motion.


nicholasmalaya:
This is a good catch :+1:

pbauman:
Is there a regression test that should get a similar update? If not, feel free to merge.

roystgnr:
Nope; the regression tests all run 2D pseudofans.  No opportunity for the same mistake.

PR #393, opened 2016-04-29
Neumann BC sign: Redo FEVariableBase setter/SolidMechanics Sign Convention

pbauman:
API for setting Neumann BC sign introduced in #386 was stupid. This one is better. Also set up the correct sign for `SolidMechanics` `Physics`.  This does not effect code in master, but rather downstream in the BC refactoring, where it has been successfully tested.

### DIFF
--- a/src/physics/include/grins/elastic_cable_base.h
+++ b/src/physics/include/grins/elastic_cable_base.h
@@ -43,14 +43,6 @@ namespace GRINS
 
   protected:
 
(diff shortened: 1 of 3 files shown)


roystgnr:
:+1:  Looks good, but if we end up adding a physics where the BC terms have to be multiplied by 1i, then egg on your face.

pbauman:
Pssh. Like there's room for any more eggs. Thanks!

PR #395, opened 2016-04-29
Doxygen Option Changes

pbauman:
Set `TEMPLATE_RELATIONS = YES` and `INLINE_INHERITED_MEMB = NO`. The former is what I originally expected when using Doxygen with CRTP-type base classes. The latter I've really come to prefer. It significantly declutters looking for methods in the current class while still easily makes available base class methods.


roystgnr:
:+1:

PR #400, opened 2016-05-16
Update Tests

pbauman:
The first updated is `regression/thermally_driven_2d_flow.sh` to use `adiabatic` for the boundary condition. This is what was there anyway, but the `general_heat_flux` bit is going away downstream, so going ahead and changing it here. There was no change to the gold data as it was the same solution.

This second `regression/sa_2d_turbulent_channel.sh` does update the gold data. Before this PR, the "v" component of velocity wasn't being pinned at the inlet. This now pins it to 0 at the inlet; the resulting change was small (5th digit), but enough to require updating the gold data.


pbauman:
Went ahead and updated more tests and got rid of the `test_thermally_driven_flow` test driver.

pbauman:
Went ahead and updated the reacting low mach tests to use the `generic_solution_regression`. Just needed to add initial conditions to the input file. Helps with fewer changes downstream in the bc-refactor branch.

pbauman:
I'd like to merge this today as it feeds the bc-refactor branch (which I've almost completed).

pbauman:
Added one more test where we use the mole fraction specification for the Dirichlet boundary condition. Didn't have coverage of this feature.

PR #401, opened 2016-05-16
More FEVariableBase Updates

pbauman:
Added two more accessors needed. Also updated `GenericFETypeVariable` to make `section_name()` public so the `Physics` can query it when registering. Finally, now have `ConvectionDiffusion` registering it's variable with the `VariableWarehouse`.


pbauman:
I'd like to merge this today as it feeds the bc-refactor branch (which I've almost completed).

pbauman:
Added one more helper function I needed for the mole fraction boundary condition parsing (downstream).

Issue #26, opened 2013-09-13
Update BC factory build API's


pbauman:
Right now, build_\* doesn't take a `GetPot` object and it probably should.

pbauman:
Closed by #403.

PR #403, opened 2016-05-18
Boundary Condition Refactor

pbauman:
This complete refactors the internals of boundary condition construction. All of the old (gross) BCHandling stuff is gone. Instead, now we use the abstract factory pattern based on libMesh::Factory (instead of the much more verbose abstract factory pattern that was there before).

I went to quite a bit of trouble to preserve the history and make it more readable; hopefully that comes through. (There was probably > 500 commits that went into this.)

Commits ef358f0 through cdeb8c4 were for building up the different factories. Note I've use "OldStyle" to indicate the previous style of input parsing (so we can easily grep and remove). Following the factories, I build up the `DefaultBCBuilder` (handles new style of parsing/factory delegation) and `OldStyleBCBuilder` (handles old style of parsing/factory delegation). Then the final commits remove all the old cruft.

All the old boundary condition input formats should work (and I believe have test coverage), except for the following:
- `parabolic_profile` (removed because this was gross, use `parsed_dirichlet` instead)
- `gaussian` (removed because this was gross, use `parsed_dirichlet` instead)
- `general_heat_flux`/`general_velocity` (fundamentally different from new factory pattern)
- `prescribed_heat_flux` (before, specified vector (god only knows why I did that), where as scalar is expected now (for scalar variables) )

As such, there were two tests that had to be updated with the migration to the new BC construction paradigm: `regression/axi_thermally_driven_flow.sh`, `regression/sa_2d_turbulent_channel.sh`

With the above exceptions, the old format should be supported. All the tests currently pass (without modification). The new format is based around the Variables. Consider three examples: the cavity benchmark boundary condition input in the new format (this is very close to what was discussed some time ago in the [grins-revel thread](https://groups.google.com/forum/#!topic/grins-devel/AGVvODmfzh8)), one of the periodic tests, and the backward facing step:

regression/low_mach_cavity_benchmark.sh
<pre>
[BoundaryConditions]
   bc_ids = '0:2 1 3'
   bc_id_name_map = 'AdiabaticWalls ColdWall HotWall'

   [./AdiabaticWalls]
      [./Velocity]
         type = 'no_slip'
      [../]
      [./Temperature]
         type = 'adiabatic'
      [../]
   [../]

   [./ColdWall]
      [./Velocity]
         type = 'no_slip'
      [../]
      [./Temperature]
         type = 'isothermal'
         T = '${* ${Materials/TestMaterial/ReferenceTemperature/value} ${- 1.0 ${Physics/LowMachNavierStokes/epsilon} } }'
      [../]
   [../]

   [./HotWall]
      [./Velocity]
         type = 'no_slip'
      [../]
      [./Temperature]
         type = 'isothermal'
         T = '${* ${Materials/TestMaterial/ReferenceTemperature/value} ${+ 1.0 ${Physics/LowMachNavierStokes/epsilon} } }'
      [../]
   [../]
[]
</pre>

exact_soln/poisson_periodic_2d_x.sh
<pre>
[BoundaryConditions]                                                                                                                                                                                                                          
   bc_ids = '0:2 1:3'
   bc_id_name_map = 'Dirichlet Periodic'

   [./Dirichlet]
      [./Temperature]
         type = 'parsed_dirichlet'
         u = '0.0'
      [../]
   [../]
   [./Periodic]
      type = 'periodic'
      master_id = '1'
      slave_id = '3'
      boundary_offset = '-1.0 0.0'
   [../]
[]
</pre>

regression/backward_facing_step.sh
<pre>
[BoundaryConditions]
   bc_ids = '1 2 3'
   bc_id_name_map = 'Inlet Walls Outlet'

   [./Inlet]
      [./Velocity]
         type = 'parsed_dirichlet'
         u = '240.0*y*(1-2*y)'
         [../]
      [../]
   [./Walls]
      [./Velocity]
         type = 'no_slip'
         [../]
      [../]
   [./Outlet]
      [./Velocity]
         type = 'homogeneous_neumann'
[]
</pre>

- BoundaryCondition gets its own section
- The `bc_ids` and `bc_id_name_map` give the look up for each boundary. The names are arbitrary, user-specified. (This was inspired a bit by FIN-S.) Using the `:`, multiple boundaries can be lumped into the same boundary. E.g. above `AdiabaticWalls` refers to boundary ids 0 and 2.
- Within each boundary section, the boundary condition is specified for each Variable (that is not a "constraint" variable, e.g. Pressure for the flow Physics).
- There are two cases where 'type` is allowed at the boundary name level: `axisymmetric` and `periodic` (I'll show another example below).
- When specifying Dirichlet type BCs (that are not things like `no_slip`), the variable name(s) (based on what was set for the names in the Variables) is what is searched for. For Variables with multiple components, e.g. Velocity, if the variable is not specified, then it is assumed to be zero. I think this is sensible behavior, and it facilitates easy specification of, for example, specifying Dirichlet boundary conditions for chemical mixtures with hundreds of components. (Before, we would've have to put in the hundreds of 0s.)
- When specifying Neumann type BCs, the variable `normal_flux` is searched for (or `traction` for solid mechanics). The size of the vector must coincide with the number of active components of the corresponding Variable.

Although this is probably more verbose in general, I think this is *much* more readable. Further, this allows a lot more error checking. In particular, there's an error if not every non-constraint variable has a boundary condition applied for every boundary (except for `axisymmetric` and `periodic` types). Thus, for "do nothing" boundaries, we'll need to actually put in the `type = homogeneous_neumann`. There's nothing actually added to the system, but it hopefully will help reduce the number of errors in the input file.

The only thing left to do is to squash together a few of the last commits, flip all the test and example input files to the new format, and do one last documentation pass. I'll do that in the morning, but I wanted to get this posted now so folks can attempt to look at it.

Where there is a net positive line count (sadly) this is actually supporting two different input file formats.

I'll let this sit for comments until 5/25 at which point I'll merge.


pbauman:
OK, one #405 is merged, I'll rebase this, but otherwise, everything is in as far as I'm concerned. All  test and example input files converted and I dumped descriptions in the master example input file in  	871e0cb. Will be very glad to have this merged and in the rear view mirror.

pbauman:
Closes #26.

pbauman:
Updated master input based on @roystgnr comments. Also realized I never actually registered `homogeneous_dirichlet` so I squashed in a fixup comment on `HomogeneousDiricheltBCFactory`.

pbauman:
Rebased on #405. Also squashed a fix into `ConstantFunctionDirichletBCFactory` that emerged when running in dbg mode with all the `GLIBCXX_DEBUG` flags turned on (mismatched iterator comparison). dbg mode `make check` runs clean now. I'd like to merge this COB tomorrow, 5/25.

pbauman:
Just a reminder that I plan to merge this before I go to bed tonight barring any comments etc.

Issue #406, opened 2016-05-20
[Feedback]: Subdomain-Restricted Variables Implementation Plan


pbauman:
We have an urgent need to support this capability (for FSI, but it will be neat to also easily be able to do conjugate heat transfer). I've been thinking about how to do this and I wanted to outline my plan here for feedback (parts that are a bad idea, things I missed, etc.).

First, this will not be possible with the "old style" of variable input (in the Physics section), but in this plan, that style will still be preserved, but it won't get the new capability.

I was thinking the input should look like the following:

<pre>
[Variables]
   [./Displacement]
      names = 'Ux Uy'
      order = 'FIRST'
      fe_family = 'LAGRANGE'
      enabled_subdomains = '1 2'
   [../]
[]
</pre>

The `enabled_subdomains` would be optional - if it's not present, then it is assumed that the Variable is enabled for all subdomains. 

For construction of Variables, I was thinking that instead of having the Physics construct them (as it is now), I can make a `VariableBuilder` that uses a `VariableFactory` that handles the construction of Variable objects and registers them with the `VariableWarehouse`. At Variable construction time, before the `Physics` enters the picture, we can already have the names, order, type, and subdomains parsed and added to the `System`. A nice side-effect of this is that it will be much easier for a user to add their own Variable.

Then, the Physics just queries the `VariableWarehouse`. This will allow error checking for things like making sure the Variable is enabled for all the subdomains for which the `Physics` is enabled. Then, the Physics can still internally set the `is_constraint_var()`, `neumann_bc_sign`, and (to be implemented) `is_time_varying()` (so we can remove the `set_time_evolving_vars()` out of the `Physics`) using setter functions.

The `VariableBuilder` would be similar style to what I did in #403 where there's the "OldStyle" (which would just be extracting the bits from the current `FEVariableBase` hierarchy) and the `DefaultVariableBuilder`. The benefit of this, beyond getting Variable construction logic out of the `Physics` and not having duplicate Variable objects floating around, is we make the `FEVariableBase` objects more lightweight: the Builder and Factories worry about construction so we only put what the `Physics` and boundary conditions need to access in the `FEVariableBase`.

In particular, I think `FEVariableBase` should only need to provide access to: variable indices, variable names, subdomain ids, material (actually there would be a map between the material and subdomain ids); I need the material in the catalytic wall boundary conditions to pick out the thermochemistry library type. 

[...]

PR #407, opened 2016-05-26
Variable Construction Refactoring

pbauman:
This is phase 1 of #406. This will break input backward compatibility for users of `ScalarODE` and `AveragedTurbineBase` subclasses. As such, this shouldn't be merged until a release has been tagged (following the merging of #403).

I also still need to update the examples and the master input file.

This moves the Variable construction to the Builder+Abstract Factory Pattern. It also now allows arbitrary naming of relevant variables. This is accomplished by adding a `type` argument to each `Variable` subsection (as suggested by @roystgnr in #406). The `type` argument is optional if the `Variable` names correspond to the `type`. This way, you can just use the type names as the `Variable` names and not actually have to specify the variable names in the Physics section. Otherwise, you will need to specify the variable names in the `Physics` section and then the corresponding `type` in the `Variable` subsection. Note this effects the BoundaryConditions as well, since they're tied to the `Variables` names.

This will facilitate subdomain-restricted variables.

`ScalarODE` and `AveragedTurbineBase` are backward incompatible because they locally stored and hardcoded their variable. Thus, no input was required to specify the `SCALAR` variable used there. This will now require the input specification for that variable (not the actual `SCALAR` part, but the other components, see `ScalarVariableFactory`).

With the exception of `type` as discussed above (and `fe_family` for the `ScalarVariable`/`ThermoPressureVariable`), all arguments are required: `fe_family`, `order`, and `names`.

Furthermore, for the vector variables (`Velocity` and `Displacement`), my intent is that the number of components of `names` will indicate the dimension of the problem to be solved (instead of the mesh dimension). This hasn't been deployed here yet, but I intended to.

So, for example, if you wanted to name the Variables something different,

<pre>
[Physics]
   [./IncompressibleNavierStokes]
      velocity_var_name = '2DVelocity'
      pressure_var_name = 'HydrostaticPressure'
   [../]
[]
[Variables]
   [./2DVelocity]
      type = 'Velocity'
      fe_family = 'LAGRANGE'
      order = 'FIRST'
      names = 'Ux Uy'
   [../]
   [./HydrostaticPressure]
      type = 'Pressure'
      ...
   [../]
[]
[BoundaryConditions]
   bc_ids = '1'
   bc_id_name_map = 'Boundary'
   [./Boundary]
      [./2DVelocity]
         ...
      [../]
   [../]
[]
</pre>


pbauman:
Oh, @roystgnr, I also meant to mention that I actually deployed my own replacement of `libMesh::Factory` in 0f3f1e5 in order to use the `get_factory()` method to reduce code duplication. I'm happy to push that back to libMesh, if you like.

pbauman:
Rebased on master after the 0.7.0 release.

Things still to do:
* Update master example input file
* Update any examples that don't have this format
* Update `Physics` that use `Velocity` or `Displacements` variables to use the `dim()` of the `Variables` to determine how many components should be computed. Separate PR?
* Update restarting to build up variables based on the restart file. This will be in a separate PR since we need to update the the XdrIO on the `libMesh` to be able to just parse the header information without building up the `System`.

pbauman:
OK, check the examples and they all use the `Variables` section bits. I added discussion to the master example input file in a289fab. I'd like to merge this ASAP. I'll do separate PRs for the last two bullet points discussed above (and I'll open tickets to track them). @nicholasmalaya, I recommend you use the 0.7.0 release until I (or someone else...) can implement the mechanism to parse the restart file for the variable ordering so that you can still use your current restart files. After this, as discussed in #406, the variable ordering in the `System` now is dictated by the order of the variables in the `[Variables]` section so it's possible (likely) that after this PR is merged, restarting from existing restarts would be hosed.

pbauman:
OK, I'm merging this. Again, @nicholasmalaya (or anyone currently with a set of restarts they're using), use the 0.7.0 release until we update the restart mechanism.

PR #409, opened 2016-05-30
Support Subdomain-Restricted Variables

pbauman:
This builds on #407. Will rebase this once it's merged. This adds data structure, parsing, and error-checking support for subdomain-restricted variables. This includes a simple conjugate heat transfer example of a "hot" cylinder in a colder flow; that example is used as a regression test.


pbauman:
Noting here that for this example to work as-is, the libMesh installation must include the fix in libMesh/libmesh#972. (Note that there is a work-around for this specific example that will work: you can make the `Temperature` variables first in the list. That won't necessarily work in general on subdomain-restricted variable problems though.)

pbauman:
Actually, I still had that `Temperature` variable at the top of the list in both the example and the test. I left it there in the example, but moved it back in the test so we actually hit libMesh/libmesh#972.

pbauman:
This will close #409.

pbauman:
OK, this PR is ready to be merged. I rebased after merging #407 (@borisboutkov, this means I've hosed you for rebasing your IBM branch. I will help you.), updated the master input file, and removed some files that should've been removed in #407. This should be much easier to review now, so I'll leave this up until Weds. night (6/22) before merging.

PR #417, opened 2016-06-30
Use modern libMesh format for SA grid

roystgnr:
Otherwise we fail a regression test with libMesh after
https://github.com/libMesh/libmesh/pull/1013


pbauman:
lulz

PR #418, opened 2016-07-05
RayfireMesh class

tradowsk:
Adds the RayfireMesh class to QoI. This class creates a 1D mesh representing a straight path across a mesh. The intention is to use this to integrate functions along the rayfire.
Currently, only 2D meshes are supported. The CPPUnit test suite uses QUAD4 and QUAD9 element types, though any element type where build_side() returns an EDGE should (in theory) work.
More element types will be tested and supported as need arises.

### DIFF
--- a/src/qoi/include/grins/rayfire_mesh.h
+++ b/src/qoi/include/grins/rayfire_mesh.h
@@ -1,0 +1,126 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// GRINS - General Reacting Incompressible Navier-Stokes
+//
+// Copyright (C) 2014-2016 Paul T. Bauman, Roy H. Stogner
+// Copyright (C) 2010-2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+
+#ifndef GRINS_RAYFIRE_MESH_H
+#define GRINS_RAYFIRE_MESH_H
+
+// libMesh
+#include "libmesh/quadrature.h"
+#include "libmesh/mesh.h"
+
+// GRINS
+#include "grins/qoi_base.h"
+#include "grins/variable_name_defaults.h"
+
+namespace GRINS
+{
+    
+  //! RayfireMesh
+  /*!
+  This class creates a 1D mesh of EDGE2 elements to allow
+  for integration of functions along a line across the domain
+  */
+  class RayfireMesh
+  {
+  public:
+    
+    //! 2D Constructor
+    /*! 
+    The init() function must be called after the constructor
+    @param origin Origin point (x,y) of the rayfire on mesh boundary
+    @param theta Spherical polar angle (in radians)
+    */
+    RayfireMesh(libMesh::Point& origin, libMesh::Real theta);
+    
+    //! 3D Constructor
+    /*! 
+    The init() function must be called after the constructor
+    @param origin Origin point (x,y,z) of the rayfire on mesh boundary
+    @param theta  Spherical polar angle (in radians)
+    @param phi    Spherical azimuthal angle (in radians)
+    */
+    RayfireMesh(libMesh::Point& origin, libMesh::Real theta, libMesh::Real phi);
+    
+    //! Initialization
+    /*! 
+    This function performs the rayfire and assembles the 1D mesh 
+    @param mesh_base Reference to the main mesh
+    */
+    void init(const libMesh::MeshBase& mesh_base);
+    
+    /*! 
+    This function takes in an elem_id on the main mesh and returns an elem from the 1D rayfire mesh 
+    @param elem_id The ID of the elem on the main mesh
+    */
+    const libMesh::Elem* translate(const libMesh::dof_id_type elem_id);
+
+
+  private:
+    //! Dimension of the main mesh 
+    const unsigned int _dim;
+    
+    //! Origin point
+    libMesh::Point& _origin;
+    
+    //! Rayfire Spherical polar angle (in radians)
+    libMesh::Real   _theta;
+    
+    //! Rayfire Spherical azimuthal angle (in radians)
+    libMesh::Real   _phi;
+    
+    //! Internal 1D mesh of EDGE2 elements
+    SharedPtr<libMesh::Mesh> _mesh;
+    
+    //! Map of main mesh elem_id to rayfire mesh elems
+    std::map<libMesh::dof_id_type,libMesh::Elem*> _elem_id_map;
+    
+    
+    //! Calculate the intersection point
+    /*!
+    @param[out] end_point The intersection point in (x,y) coordinates
+    @return Elem* if intersection is found
+    @return NULL  if no intersection point found (i.e. start_point is on a boundary)
+    */
+    const libMesh::Elem* _get_next_elem(const libMesh::Elem* cur_elem, libMesh::Point* start_point, libMesh::Point* end_point);
+    
+    //! Ensure the calculated intersection point is on the edge_elem and is not the start_point
+    bool _check_valid_point(libMesh::Point& intersection_point, libMesh::Point& start_point, libMesh::Elem& edge_elem, libMesh::Point* next_point);
+    
+    //! Knowing the intersection_point, get the appropraite next elem along the path
+    const libMesh::Elem* _get_correct_neighbor(libMesh::Point& end_point, const libMesh::Elem* cur_elem, unsigned int side);
+    
+    //! Ensure the supplied origin is on a boundary of the mesh
+    void _check_origin_on_boundary(const libMesh::Elem* start_elem);
+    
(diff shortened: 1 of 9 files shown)


nicholasmalaya:
Just throwing this out, but I have no idea what a Rayfire Mesh is. Might be worth throwing a reference to what it is into the src, likely in 'rayfire_mesh.h'. I noticed you had a block touching on this, 

<pre>
//! RayfireMesh
/*!
   This class creates a 1D mesh of EDGE2 elements to allow
    for integration of functions along a line across the domain
</pre>

But that isn't really enough for someone such as myself to understand what this is. Even just a hyper-link in the comment block to a paper this is based on would be worthwhile, IMO.

pbauman:
@nicholasmalaya, thanks for the feedback. This is simply doing a ray fire from a given boundary point and direction across the mesh; no paper or anything this is based on. The ray fire will cache each element that it crosses into a libMesh::MeshBase. The ultimate goal is to integrate quantities along this line for a given solution field; we store everything as a 1D mesh so we can more easily handle this QoI under AMR without doing the full ray fire again.

@tradowsk, definitely need more documentation about the class in the location where @nicholasmalaya pointed out. Spell out what you're doing (e.g. given point, direction, walk across the mesh by... etc. etc.).

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -0,0 +1,126 @@
...
+    This function performs the rayfire and assembles the 1D mesh 
+    @param mesh_base Reference to the main mesh
+    */
+    void init(const libMesh::MeshBase& mesh_base);
+    
+    /*! 
+    This function takes in an elem_id on the main mesh and returns an elem from the 1D rayfire mesh 
+    @param elem_id The ID of the elem on the main mesh
+    */
+    const libMesh::Elem* translate(const libMesh::dof_id_type elem_id);
```
pbauman: This is not a good function name for this. `translate` to me means translating an element or something. I'd do something like `map_main_elem_to_1d_elem()`. Also, how does this function behave if there's no underlying 1D element? The documentation should tell the user.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -0,0 +1,126 @@
...
+    std::map<libMesh::dof_id_type,libMesh::Elem*> _elem_id_map;
+    
+    
+    //! Calculate the intersection point
+    /*!
+    @param[out] end_point The intersection point in (x,y) coordinates
+    @return Elem* if intersection is found
+    @return NULL  if no intersection point found (i.e. start_point is on a boundary)
+    */
+    const libMesh::Elem* _get_next_elem(const libMesh::Elem* cur_elem, libMesh::Point* start_point, libMesh::Point* end_point);
```
pbauman: The convention we use is only class variables get a `_` prefix in the name; functions, even private ones, do not get a `_` prefix.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -0,0 +1,126 @@
...
+  {
+  public:
+    
+    //! 2D Constructor
+    /*! 
+    The init() function must be called after the constructor
+    @param origin Origin point (x,y) of the rayfire on mesh boundary
+    @param theta Spherical polar angle (in radians)
+    */
+    RayfireMesh(libMesh::Point& origin, libMesh::Real theta);
```
pbauman: I would more explicitly say this function is restricted to 2D meshes. (Yes, I see you said 2D constructor and yes I see you said (x,y), but I think the extra explicitness is worth it to remove any ambiguity.)

src/qoi/src/rayfire_mesh.C
```diff
@@ -0,0 +1,331 @@
...
+#include "libmesh/edge_edge2.h"
+#include "libmesh/analytic_function.h"
+#include "libmesh/enum_elem_type.h"
+#include "libmesh/fe.h"
+#include "libmesh/fe_interface.h"
+
+namespace GRINS
+{
+  RayfireMesh::RayfireMesh(libMesh::Point& origin, libMesh::Real theta, libMesh::Real phi) : 
+    _dim(3),_origin(origin),_theta(theta),_phi(phi)
```
pbauman: Each of these goes on their own line, please.

src/qoi/src/rayfire_mesh.C
```diff
@@ -0,0 +1,331 @@
...
+  {
+    libmesh_not_implemented();
+  }
+ 
+ 
+  RayfireMesh::RayfireMesh(libMesh::Point& origin, libMesh::Real theta) : 
+    _dim(2),_origin(origin),_theta(theta),_phi(0)
```
pbauman: For the 2D case, initialize _phi to some nonsensical value instead of 0 so you can assert later that phi was set to a valid value.

pbauman: You need to check that `theta` is in a valid range and `libmesh_error_msg` if it's not.

src/qoi/src/rayfire_mesh.C
```diff
@@ -0,0 +1,331 @@
...
+  bool RayfireMesh::_check_valid_point(libMesh::Point& intersection_point, libMesh::Point& start_point, libMesh::Elem& edge_elem, libMesh::Point* next_point)
+  {
+    bool is_not_start = !(intersection_point.absolute_fuzzy_equals(start_point));
+    bool is_on_edge = edge_elem.contains_point(intersection_point);
+        
+    if ( is_not_start && is_on_edge )
+    {
+      (*next_point)(0) = intersection_point(0);
+      (*next_point)(1) = intersection_point(1);
+      return true;
```
pbauman: I'm not a big fan of multiple `returns` in a function if it's easy to avoid it.

src/qoi/src/rayfire_mesh.C
```diff
@@ -0,0 +1,331 @@
...
+        
+        if (elem == cur_elem) // skip the current elem
+          continue;
+        
+        // move a little bit along the rayfire
+        // and see if we are in the elem
+        libMesh::Real L = elem->hmin();
+        L *= 0.1;
+        
+
```
pbauman: One line of whitespace is enough.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -0,0 +1,126 @@
...
+    
+    //! Ensure the supplied origin is on a boundary of the mesh
+    void _check_origin_on_boundary(const libMesh::Elem* start_elem);
+    
+    //! Iterative solver for calculating the intersection point of the rayfire on the side of an elem
+    /*!
+    @param[out] intersection_point The calculated intersection point (only set if convergence is achieved)
+    @return Whether or not the solver converged before hitting the iteration limit
+    */
+    bool _newton_solver(libMesh::Point& initial_point, const libMesh::Elem* edge_elem, libMesh::Point* intersection_point);
```
pbauman: I'd like more documentation here. Specifically, what is the residual function (i.e. the math you did on the whiteboard), the Jacobian, parameters, etc. for the Newton solve.

test/unit/rayfire_test.C
```diff
@@ -0,0 +1,365 @@
...
+    CPPUNIT_TEST( test_quad4_2D );
+    CPPUNIT_TEST( test_quad9_2D );
+    CPPUNIT_TEST( fire_through_vertex );
+
+    CPPUNIT_TEST_SUITE_END();
+
+  public:
+  
+    // these tests should produce a "bad origin" error
+/*    void origin_not_on_boundary()
```
pbauman: Let's go ahead and trim this out.

src/qoi/src/rayfire_mesh.C
```diff
@@ -0,0 +1,331 @@
...
+ 
+ 
+  RayfireMesh::RayfireMesh(libMesh::Point& origin, libMesh::Real theta) : 
+    _dim(2),_origin(origin),_theta(theta),_phi(0)
+  {}
+
+
+  void RayfireMesh::init(const libMesh::MeshBase& mesh_base)
+  {
+    _mesh = new libMesh::Mesh(mesh_base.comm(),(unsigned char)1);
```
pbauman: You should check here that you have consistency: if the user said this was two-dimensional, you should make sure you have a two-dimensional thing and that the 3D angle hasn't been set.

tradowsk:
Sorry for the delay, was having doxygen issues that I could not straighten out remotely.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -39,24 +39,40 @@ namespace GRINS
     
   //! RayfireMesh
   /*!
-  This class creates a 1D mesh of EDGE2 elements to allow
-  for integration of functions along a line across the domain
+  This class performs a rayfire across a given mesh.
+  The user supplies an origin point on the mesh boundary,
+  and spherical angle(s) theta (and phi for 3D) to describe the direction of the rayfire.
```
pbauman: \theta is the angle from the x-axis in the x-y plane? \phi is the angle from the x-y plane with positive being in the positive half of the z-plane? If so, please indicate. Otherwise, please specify what the angles are.

pbauman:
Much improved, thanks. One last clarification and I think this will be ready to merge.

tradowsk:
@pbauman I squashed in the doxygen clarification of the angles. Anything else?

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -39,24 +39,44 @@ namespace GRINS
...
+  As is standard in spherical coordinates, theta is the angle from the x-axis in the xy-plane,
+  with counterclockwise being positive. -2&pi; &le; theta &le; +2&pi;.
+  Phi is the angle from the xy-plane, with phi>0 being in the positive half of the z-plane.
+  -2&pi; &le; phi &le; +2&pi;
```
pbauman: phi goes from -\pi to \pi, right?

Also, OK for now, but for future info, you can enclose LaTeX statements in Doxygen comments with `\f$ <latex> \f$` or, for a full equation, `\f[ <latex> \f]`.

tradowsk: oops, nice catch. It has been fixed.

pbauman:
Thanks!

PR #419, opened 2016-07-07
Use newest tsuna/boost.m4

roystgnr:
This fixes a configuration error with icpc 16.0 for me; it apparently
was added to fix the same error triggered by gcc 6.1.1:

https://github.com/tsuna/boost.m4/commit/5c9afc0e7bc329c8b5431a87b24ae82a04c6e756


pbauman:
Thanks!

PR #420, opened 2016-07-08
Reacting Low Mach SPGSM Stabilization

pbauman:
This PR adds "SPGSM" stabilization (SUPG+PSPG+DivGrad) for Reacting Low Mach Navier Stokes. Currently, we have an Ozone flame example, but with constant transport models. I used the transient solution that was close enough to steady-state that gets a steady state solve for a regression test. I'm going to add an XFAIL test for the currently not working Antioch "KineticsTheory" transport before merging so we can track that as we work out the issue. In the mean time, @roystgnr and @hovr2pi you can play with the example.


roystgnr:
Still planning to add the XFAIL test soon?  If not then we could just merge this as-is.

pbauman:
Thanks for the prod. Rebased and added the XFAIL test.

PR #422, opened 2016-07-19
Add preassembly() hook for Physics

pbauman:
We override `FEMSystem::assembly()` and first do a `preassembly()` pass over each Physics. I have at least two use cases for this.

The one thing that concerns me though is that `System::update()` is called in `FEMSystem::assembly()` and one my of use cases will need to use the solution (to accumulate solution values along the (1D) mesh). This means that we'll endup calling `System::update()` twice in that case.

One solution could be we add an optional fourth argument to `FEMSystem::assembly()` - `bool do_solution_localization = true`. This will be compatible with internal library usage and give the hook for us to do the `update()` first and avoid the redundancy, but will obviously break API compatibility for anyone overriding `FEMSystem::assembly()`. Thoughts @roystgnr? (Obviously, this last bit is orthogonal to this PR, but wanted to bring it up for discussion in this context before cooking up the libMesh PR.)

### DIFF
--- a/src/physics/include/grins/multiphysics_sys.h
+++ b/src/physics/include/grins/multiphysics_sys.h
@@ -124,6 +124,12 @@ namespace GRINS
     //! Context initialization. Calls each physics implementation of init_context()
     virtual void init_context( libMesh::DiffContext &context );
 
+    //! Override FEMSystem::assembly
+    /*! This allows us to insert things like preassembly(). */
+    virtual void assembly( bool get_residual,
+                           bool get_jacobian,
+                           bool apply_heterogeneous_constraints = false );
+
     // residual and jacobian calculations
     // element_*, side_* as *time_derivative, *constraint, *mass_residual
 
--- a/src/physics/include/grins/physics.h
+++ b/src/physics/include/grins/physics.h
@@ -162,6 +162,9 @@ namespace GRINS
     //! Initialize context for added physics variables
     virtual void init_context( AssemblyContext& context );
 
+    //! Perform any necessary setup before element assembly begins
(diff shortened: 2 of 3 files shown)


roystgnr:
:+1: for this change.

Adding an additional flag to assembly() arguments isn't a good option; that'll break the API for anyone else who's decided to override.

Adding an "update_in_assembly" flag to FEMSystem would be acceptable, but is it really necessary?  This feels like premature optimization.

pbauman:
> Adding an "update_in_assembly" flag to FEMSystem would be acceptable, but is it really necessary? This feels like premature optimization.

Yeah, you're probably right. If/when it becomes a performance, we can address it then. Thanks.

roystgnr:
@nicholasmalaya posted http://number-none.com/blow/john_carmack_on_inlined_code.html to chat today; I'm about 90% sure that this was a coincidence, but when I reached

> The way we have traditionally measured performance and optimized our games encouraged a lot of conditional operations -- recognizing that a particular operation doesn't need to be done in some subset of the operating states, and skipping it. This gives better demo timing numbers, but a huge amount of bugs are generated because skipping the expensive operation also usually skips some other state updating that turns out to be needed elsewhere.

> We definitely still have tasks that are performance intensive enough to need optimization, but the style gets applied as a matter of course in many cases where a performance benefit is negligible, but we still eat the bugs.

it sure felt like he was making a passive-aggressive crack about your suggestion in this ticket.  ;-)

pbauman:
Damn. Rekt.

pbauman:
Rebased.

PR #423, opened 2016-07-19
RayfireMesh AMR

tradowsk:
This PR adds AMR support to the RayfireMesh class (added in #418) and associated CPPUnit tests.

Refinement of the Rayfire mesh is done through the `RayfireMesh::reinit()` function, which must be called after each refinement of the main mesh. Right now, this must be done manually. However, in the near future, we will hook into MultiphysicsSystem to call `reinit()` on RayfireMesh objects in an upcoming QoI class.

The API change in the first commit was necessary to allow `RayfireMesh::reinit()` to have an empty parameter list (as will be required when the aforementioned hooks are added). Without a MeshBase object like in the `RayfireMesh::init(MeshBase&)` function, there was no was to get an `Elem*` from the elem IDs stored in the `RayfireMesh::_elem_map`. By storing `Elem*` instead of just their IDs, we can reinit without needing a MeshBase.

`make check` passes using latest `libmesh/master` with GCC 4.8.4 and 5.3.0

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -135,6 +135,32 @@ namespace GRINS
 
     return NULL;
   }
+  
+  
+  void RayfireMesh::reinit(const libMesh::MeshBase& mesh_base)
+  {
+    // store the elems to be refined until later
+    // so we don't mess with the _elem_id_map while we
+    // iterate over it
+    std::vector<std::pair<const libMesh::Elem*,libMesh::Elem*> > elems_to_refine;
+
+    // iterate over all elems along the rayfire and looks for INATCIVE ones
+    // that were just refined
+    std::map<libMesh::dof_id_type,libMesh::Elem*>::iterator it = _elem_id_map.begin();
+    for(; it != _elem_id_map.end(); it++)
+      {
+        const libMesh::Elem* main_elem = mesh_base.elem(it->first);
+        libmesh_assert(main_elem);
+        libMesh::Elem::RefinementState state = main_elem->refinement_flag();
+
+        if(state == libMesh::Elem::INACTIVE)
+          elems_to_refine.push_back(std::pair<const libMesh::Elem*, libMesh::Elem*>(main_elem,it->second));
+      }
+
+    // refine the elements that need it
+    for (unsigned int i=0; i<elems_to_refine.size(); i++)
+      refine(elems_to_refine[i].first, elems_to_refine[i].second);
+  }
 
 
   // private functions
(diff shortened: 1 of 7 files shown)


pbauman:
A few comments.

1. For new things where there hasn't been a release, e.g. `RayfireMesh`, API changes are OK.
2. Why does `RayfireMesh::reinit()` need to be empty? You can mandate what you need in the API, particularly if it facilitates cleaner code. In this case, you should be able to have `MultiphysicsSystem` available to you (which would give you everything), but I think you only need the mesh in this case?
3. Given point 2, I think the internal change is unnecessary and I prefer the element ids for clarity.
4. I see at least 3 separate commits that could be split out of 131a2a2: whitespace, `reinit()`, and the CppUnit testing.

tradowsk:
>2.Why does  RayfireMesh::reinit()  need to be empty? You can mandate what you need in the API, particularly if it facilitates cleaner code. In this case, you should be able to have  MultiphysicsSystem  available to you (which would give you everything), but I think you only need the mesh in this case?
3.Given point 2, I think the internal change is unnecessary and I prefer the element ids for clarity.

I went with this change mainly because I thought passing around a MeshBase& through that whole reinit() chain seemed a bit cumbersome. I do see your point about the loss of clarity though. Either way doesn't really make the Rayfire code itself any simpler or more convoluted, so if you think that storing the IDs is better, I can switch it back easily.

pbauman:
`libMesh::System::reinit()` and cousins are empty since they are the `System`. I think what will make sense is having the `CompositeQoi::reinit()` take a `MultiphysicsSystem` and then the `RayfireMesh::reinit()` can take a MeshBase (from the system handed to the `CompositeQoI`).

tradowsk:
That's a happy medium I think.

So how do you want these changes commit-wise? I can just do a `git reset HEAD~2` and do individual commits from there, then force push to this branch.

pbauman:
> I can just do a git reset HEAD~2 and do individual commits from there, then force push to this branch.

👍

tradowsk:
New commits address the above points. API has returned to store elem_id's in the map as in #418, and updated `RayfireMesh::reinit()` -> `RayfireMesh::reinit(MeshBase&)`.

`unit_driver` still passing with GCC 5.3.0

src/qoi/src/rayfire_mesh.C
```diff
@@ -336,5 +362,79 @@ namespace GRINS
...
+    libMesh::Point* start_point = start_node;
+    libMesh::Point* end_point = new libMesh::Point();
+
+    const libMesh::Elem* next_elem;
+    const libMesh::Elem* prev_elem = main_elem->child(i);
+    
+    // if prev_elem is INACTIVE, then more than one refinement
+    // has taken place between reinit() calls and will
+    // break this
+    libmesh_assert( prev_elem->refinement_flag() == libMesh::Elem::RefinementState::JUST_REFINED );
```
pbauman: Can you please use `libmesh_assert_equal_to()`?

src/qoi/src/rayfire_mesh.C
```diff
@@ -336,5 +362,79 @@ namespace GRINS
...
+
+    // calculate the end point and
+    // get the second elem in the rayfire
+    next_elem = get_next_elem(prev_elem,start_point,end_point);
+
+    // iterate until we reach the stored end_node
+    while(!(end_point->absolute_fuzzy_equals(*end_node)))
+      {
+        // again, checking for multiple refinements
+        libmesh_assert( next_elem->refinement_flag() == libMesh::Elem::RefinementState::JUST_REFINED );
```
pbauman: Same here. `libmesh_assert()` should be used for things like checking the validity of a pointer. Otherwise, there are assert macros for conditional statements that print more informative messages.

pbauman:
A couple of small nitpicks that you can squash in. 

Have you tested coarsening? If not, could you please add a test where you coarsen? Thanks.

tradowsk:
> Have you tested coarsening?
I did not address coarsening yet, as I thought you wanted a different PR for that. I think I know how to do it, so you want me to just add it to this PR then?

pbauman:
Oh, I see. In that case, I would update the class documentation to specifically state that the refinement capability is still experimental and that coarsening is not yet supported and coarsening can be in another PR.

tradowsk:
Squashed in those `assert` fixes as well

pbauman:
Thanks!

PR #424, opened 2016-07-20
Include serial_mesh.h in new rayfire test

roystgnr:
Otherwise when we build with a --enable-parmesh libMesh, this test
doesn't know what a serial mesh is.


pbauman:
Thanks. Hopefully Civet will be released soon to catch stuff like this.

PR #425, opened 2016-07-20
Handle boundary_info on distributed meshes

roystgnr:
This fixes errors where we were assuming that BoundaryInfo was
returning global info but it was actually returning only local info.

Apologies to @pbauman for the poor API documentation there, which
should be fixed by https://github.com/libMesh/libmesh/pull/1032

This doesn't get "make check" passing for me yet, but it fixes 90+% of the failures I was seeing initially.


pbauman:
At least all the code was in one place now! Thanks for the fix and sorry for the trouble.

PR #428, opened 2016-07-22
Rayfire coarsening

tradowsk:
Adds coarsening functionality to RayfireMesh, and a CPPUnit test for it.

Coarsening of the rayfire is done through storing the parent of refined rayfire elements. If that same element is later coarsened, we simply inactivate the children and activate the parent. The `RefinementState` is checked  in `map_to_rayfire_elem()` to ensure that only active rayfire elements get returned.

Note that the initial rayfire performed by `RayfireMesh::init()` must be done on the coarsest mesh before any refinements are done. Similar to how libMesh operates, we cannot coarsen the rayfire mesh beyond its initial version. Thus, a check was added to `RayfireMesh::init()` in 5370c06 to ensure that all elements along the rayfire are level 0.

Hooks to call `RayfireMesh::reinit()` post-refinement/coarsening will be placed into MultiphysicsSystem soon, but without a rayfire-based QoI implemented, there is no way to test them.

`make check` passes with libMesh `--enable-parmesh`

### DIFF
--- a/src/qoi/include/grins/rayfire_mesh.h
+++ b/src/qoi/include/grins/rayfire_mesh.h
@@ -59,9 +59,8 @@ namespace GRINS
   can access the 1D elements, prescribe a desired quadrature rule, and perform
   integration directly along the line, rather than using the entire 2D/3D elements
   of the main mesh.
-  
-  Refinement of the rayfire mesh is supported through the reinit() function.
-  Coarsening is not yet supported.
+
+  Refinement and coarsening of the rayfire mesh are supported through the reinit() function.
   */
   class RayfireMesh
   {
@@ -90,12 +89,13 @@ namespace GRINS
     /*!
     This function performs the rayfire and assembles the 1D mesh.
     Call this immediately after the constructor.
+    Must be called before any refinements of the main mesh.
     @param mesh_base Reference to the main mesh
     */
     void init(const libMesh::MeshBase& mesh_base);
 
     /*!
-    This function takes in an elem_id on the main mesh and returns an elem from the 1D rayfire mesh
+    This function takes in an elem_id on the main mesh and returns an elem from the 1D rayfire mesh.
     @param elem_id The ID of the elem on the main mesh
     @return Elem* to 1D elem from the rayfire mesh if the elem_id falls along the rayfire path
     @return NULL if the given elem_id does not correspond to an elem through which the rayfire passes
@@ -103,11 +103,12 @@ namespace GRINS
     const libMesh::Elem* map_to_rayfire_elem(const libMesh::dof_id_type elem_id);
 
     /*!
-    Checks for refined main mesh elements along the rayfire path.
-    If INACTIVE elements are found, they are passed to refine() to update the rayfire mesh.
+    Checks for refined and coarsened main mesh elements along the rayfire path.
+    They are then passed to refine() and coarsen(), respectively, to update the rayfire mesh.
 
-    Only 1 refinement can be done between reinit() calls.
-    Coarsening is not yet supported
+    Only 1 level of refinement and/or coarsening can be done between reinit() calls.
+    Note that this is not limited to doing either refinement or coarsening between reinits.
+    Both can be done on different elements, as long as they are only 1 level in either direction.
     @param mesh: reference to main mesh, needed to get Elem* from the stored elem_id's
     */
     void reinit(const libMesh::MeshBase& mesh_base);
@@ -132,9 +133,18 @@ namespace GRINS
     //! Map of main mesh elem_id to rayfire mesh elems
     std::map<libMesh::dof_id_type,libMesh::Elem*> _elem_id_map;
 
-
     //! User should never call the default constructor
     RayfireMesh();
+
+    //! Private function to get a rayfire elem from main_mesh elem ID
+    /*!
+    Does not return a const pointer, and is used within this class to simplify
+    rayfire elem access.
+
+    Also used by map_to_rayfire_elem(), which adds a const to prevent modification
+    outside this class.
+    */
+    libMesh::Elem* get_rayfire_elem(const libMesh::dof_id_type elem_id);
 
     //! Calculate the intersection point
     /*!
@@ -194,6 +204,10 @@ namespace GRINS
 
     //! Refinement of a rayfire element whose main mesh counterpart was refined
     void refine(const libMesh::Elem* main_elem, libMesh::Elem* rayfire_elem);
+
+    //! Coarsening of a rayfire element whose main mesh counterpart was coarsened
+    void coarsen(const libMesh::Elem* rayfire_elem);
+
   };
 
 }
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -266,7 +266,7 @@ namespace GRINS
 
     // check if the intersection point is a vertex
     bool is_vertex = false;
-    for(unsigned int n=0; n<4; n++)
+    for(unsigned int n=0; n<cur_elem->n_nodes(); n++)
       is_vertex |= (cur_elem->get_node(n))->absolute_fuzzy_equals(end_point);
 
     if (is_vertex)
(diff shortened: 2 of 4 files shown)


pbauman:
> Note that the initial rayfire performed by RayfireMesh::init() must be done on the coarsest mesh before any refinements are done. Similar to how libMesh operates, we cannot coarsen the rayfire mesh beyond its initial version. Thus, a check was added to RayfireMesh::init() in 5370c06 to ensure that all elements along the rayfire are level 0.

I don't like this. Certainly, for doing parameter sampling and such, we'll want to restart on an AMR grid that was generated before hand and this will be a major restriction. I'm not sure what the best solution is yet, either writing out the rayfire mesh and reading it back in, coarsening the "primary" mesh, doing the rayfire and then refining, walking the tree and storing all the parents, or something else, but this will be a major restriction in the future.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -59,9 +59,9 @@ namespace GRINS
   can access the 1D elements, prescribe a desired quadrature rule, and perform
   integration directly along the line, rather than using the entire 2D/3D elements
   of the main mesh.
-  
-  Refinement of the rayfire mesh is supported through the reinit() function.
-  Coarsening is not yet supported.
+
+  Refinement and coarsening of the rayfire mesh are supported through the reinit() function.
+  init() must be called before any mesh refinement is done.
```
pbauman: You need to note the restriction about starting from a mesh with no refinement.

src/qoi/include/grins/rayfire_mesh.h
```diff
@@ -132,6 +134,21 @@ namespace GRINS
     //! Map of main mesh elem_id to rayfire mesh elems
     std::map<libMesh::dof_id_type,libMesh::Elem*> _elem_id_map;
 
+    //! Map to store children of rayfire EDGE2 elems
+    /*!
+    This is necessary because it is likely that on refinement,
+    a single EDGE2 rayfire element will have 3 children
```
pbauman: "This is necessary because it is possible that upon refinement, the rayfire element (EDGE2) will have more than 2 children..."

tradowsk:
>I don't like this. Certainly, for doing parameter sampling and such, we'll want to restart on an AMR grid that was generated before hand and this will be a major restriction. I'm not sure what the best solution is yet, either writing out the rayfire mesh and reading it back in, coarsening the "primary" mesh, doing the rayfire and then refining, walking the tree and storing all the parents, or something else, but this will be a major restriction in the future.

We could also redo a portion of the rayfire in the vicinity of the coarsened element(s). Not quite as elegant or optimal as the elem-by-elem refinement, but it would get around this restriction.

src/qoi/src/rayfire_mesh.C
```diff
@@ -119,6 +116,10 @@ namespace GRINS
 
         // add new rayfire elem to the map
         _elem_id_map[prev_elem->id()] = elem;
+
+        // add new rayfire elem ID to parent_child map
+        add_parent(elem->id());
```
pbauman: Although technically valid C++, I prefer the `this->add_parent` syntax to make it clear we're calling a member function. Please adjust all other instances as well.

src/qoi/src/rayfire_mesh.C
```diff
@@ -366,75 +389,159 @@ namespace GRINS
...
+    // i.e. it was previously refined then coarsened
+    if (has_children(rayfire_elem->id()))
+    {
+      // we already have the refined children, so no need to do any calculating
 
-    // remove unrefined elem from _elem_id_map
-    std::map<libMesh::dof_id_type,libMesh::Elem*>::iterator it;
-    it = _elem_id_map.find(main_elem->id());
-    _elem_id_map.erase(it);
+      // inactivate the parent
```
pbauman: deactivate

src/qoi/src/rayfire_mesh.C
```diff
@@ -366,75 +389,159 @@ namespace GRINS
...
+
+      // set the rayfire_elem as INACTIVE
+      rayfire_elem->set_refinement_flag(libMesh::Elem::RefinementState::INACTIVE);
+
+      // find which child elem we start with
+      unsigned int i=0;
+      while(!(main_elem->child(i))->contains_point(*start_node))
+        i++;
+
+      // we found the starting element, so perform a the rayfire
```
pbauman: a the

pbauman:
OK, mostly comment/styles nitpicks. Thanks!

One more request though. Could you add a test using [this](https://github.com/grinsfem/grins/blob/master/test/grids/mixed_quad_tri_square_mesh.xda) mesh? Be sure to fire across such that the line goes through both type of elements. uniformly refining and coarsening is fine. This way, we're hitting something with both `QUAD4` and `TRI3` elements in the same mesh.

Also, have you tried refine->update rayfire->refine again? Just for giggles. You could use the previous mesh to test that too.

Thanks!

pbauman:
> We could also redo a portion of the rayfire in the vicinity of the coarsened element(s). Not quite as elegant or optimal as the elem-by-elem refinement, but it would get around this restriction.

I would be fine with this.

src/qoi/src/rayfire_mesh.C
roystgnr: Omitting this-> is only valid C++ as long as the member function isn't coming from a templated base class.  So your options for syntax, in increasing order of preference, are:
1. Never refactor your code to use a templated base class.
2. After any such refactor, spend an hour compiling, checking syntax errors, adding "this->", and repeating.
3. Just get in the habit of using "this->" all the time, even when redundant.

pbauman: @tradowsk Yes, we've been burned by this many times.

tradowsk: Good to know, will get in the habit of doing this

tradowsk:
I figured out a way to do away with the coarsest restriction without having to redo any of the rayfire. Just a few simple distance calculations.

Will update the code, docs, and tests as suggested above, and then redo the commits once everything is working

tradowsk:
Redid the coarsening functionality to remove restriction that `RafireMesh::init()` needed to be called on coarsest mesh.

Currently adding the additional tests @pbauman requested

tradowsk:
I added the additional tests to RayfireAMR:

- Test multiple successive refinements
- Test refine and coarsen on a mixed element mesh
- Test refine and coarsen mesh at the same time

pbauman:
Great, thanks. Do any of the coarsen tests exercise "rayfire happened on a mesh that wasn't at the coarsest level"?

If you need help with the upstream rebase for the conflicts, please let me know.

tradowsk:
> Do any of the coarsen tests exercise "rayfire happened on a mesh that wasn't at the coarsest level"?

Not currently, but I shall whip something up and squash it in

tradowsk:
Squashed in a test with an already refined mesh. Anything else?

test/unit/rayfireAMR_test.C
```diff
@@ -154,6 +156,143 @@ namespace GRINSTesting
...
+      rayfire->init(*mesh);
+
+      CPPUNIT_ASSERT( rayfire->map_to_rayfire_elem(mesh->elem(0)->child(0)->id()) );
+      CPPUNIT_ASSERT( rayfire->map_to_rayfire_elem(mesh->elem(0)->child(1)->id()) );
+
+      CPPUNIT_ASSERT( !(rayfire->map_to_rayfire_elem(mesh->elem(0)->child(2)->id())) );
+      CPPUNIT_ASSERT( !(rayfire->map_to_rayfire_elem(mesh->elem(0)->child(3)->id())) );
+
+      // refine again
+      mr.uniformly_refine();
```
pbauman: Following this block of checks, can you uniformly_coarsen(); rayfire->reinit; uniformly_coarsen; rayfire->reinit and make sure we're doing the right thing?

tradowsk: Squashed in

pbauman:
Great, and thanks. Onward!

PR #431, opened 2016-07-27
libMesh PR #1040 Fix

pbauman:
Updates gold data for `regression/extra_quadrature_order_laplace_arefee_amr.sh` since the refinement pattern changed. Test will now fail with libMesh prior to libMesh/libmesh#1040.

Builds on #430.


pbauman:
Merging right away.

PR #432, opened 2016-07-27
Enable AdaptiveTimeSolver with negative tolerance

roystgnr:
libMesh uses a negative value here to signal "base my tolerance on the
time discretization error from the first successful time step", and we
should pass that along to enable that feature.


pbauman:
Thanks! (and sorry for missing that)

PR #436, opened 2016-08-17
Fix regression/velocity_penalty_drag gold data

pbauman:
There was inconsistency between the data generated from the run
and the gold data in #434. Perhaps libMesh version differences?


nicholasmalaya:
Hmm, weird. This is passing locally so I think everything is fine. Good catch.

pbauman:
Thanks for checking. The only think I could think of is maybe the gold data was generated with an older libMesh that mucked with the restart ordering? Regardless, the failure was a tripped assert regarding indexing on reading the gold data file. Since this also works for you, I'll go ahead and merge.

PR #438, opened 2016-08-19
Bugfix for RayfireMesh refinement

tradowsk:
Upon refinement of rayfire elements, if the rayfire enters a parent element very near the boundary between 2 children, the starting child is not properly calculated.

A CPPUnit test was added to verify that the issue is resolved.

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -399,27 +399,43 @@ namespace GRINS
     rayfire_elem->set_refinement_flag(libMesh::Elem::RefinementState::INACTIVE);
 
     // find which child elem we start with
-    libMesh::dof_id_type start_child = -1;
+    libMesh::dof_id_type start_child = libMesh::invalid_uint;
     for(unsigned int i=0; i<main_elem->n_children(); i++)
       {
         if ( (main_elem->child(i))->contains_point(*start_node) )
           {
-            // move a little bit along the rayfire
-            // and see if we are in the elem
-            libMesh::Real L = main_elem->child(i)->hmin();
-            L *= 0.1;
+            // check if the start point is a vertex
+            bool is_vertex = false;
+            for(unsigned int n=0; n<main_elem->child(i)->n_nodes(); n++)
+              is_vertex |= (main_elem->child(i)->get_node(n))->absolute_fuzzy_equals(*start_node);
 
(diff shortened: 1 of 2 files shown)


pbauman:
Could you elaborate a bit more on the failure mode?

Also, I noticed [here](https://github.com/grinsfem/grins/pull/438/files#diff-e65e60adfe3e738c523f0053bb2a8f2fR402) that you initialize with a potentially bad value (I think that's an unsigned type). Use the corresponding `invalid_id` - grep through GRINS and/or libMesh for the right invalid_type.

tradowsk:
Squashed in use of `invalid_uint` for the initialization value.

As for the original issue, sorry for the short explanation.

When refining rayfire elements, I start with the main mesh element that was just refined. I need to figure out through which child of that main mesh element the rayfire enters. However, should that entry point or `start_point` be a vertex between 2 children, then 

`main_elem->child(i)->contains_point(start_point)` 

would not always yield a valid starting child element, since that point is contained by 2 children. So what I did originally was check if the child elem contained the point, and if so, I walked a small distance along the rayfire and checked if that point was also inside the child elem. This works great when the start_point is in the middle of the element, or on a vertex. However, when the `start_point` is very near to the vertex, but not quite on it, then my small walk along the rayfire would take me into the next child elem, and thus I was getting results where none of the children contained the `start_point` which resulted in a `segfault`.

With this fix, if a child elem contains the `start_point`, I first check if `start_point` is a vertex of the child elem. If it is, I can do my small walk to see if I am in the right child elem. If `start_point` is not a vertex, then it can only be contained by one child elem, so there is no need to do that small walk that was causing my issues.

The CPPUnit test added starts the rayfire very close to a postrefinement vertex, and after refinement, checks to make sure the right elems are included in the rayfire.

pbauman:
OK, thanks for the more detailed explanation.

pbauman:
I would add a `libmesh_assert_not_equal_to(start_child,libMesh::invalid_uint)` at the end of your child loop (looks like line 436 of rayfire_mesh.C) in case the logic every changes.

tradowsk:
Squashed in the `assert`

pbauman:
`make check` still passing? (Grumbles about Civet being open-sourced.)

tradowsk:
Yes, both for the rayfire tests and the IntegratedFunction test that originally brought up this issue

Issue #441, opened 2016-08-25
RayfireMesh Warnings


pbauman:
With GCC 6.1

<pre>
./../../../grinsfem/src/qoi/src/rayfire_mesh.C: In member function ‘void GRINS::RayfireMesh::coarsen(const libMesh::Elem*)’:
../../../../grinsfem/src/qoi/src/rayfire_mesh.C:537:44: warning: ‘end_node’ may be used uninitialized in this function [-Wmaybe-uninitialized]
         elem->set_node(1) = _mesh->node_ptr(end_node->id());
                             ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~
../../../../grinsfem/src/qoi/src/rayfire_mesh.C:536:44: warning: ‘start_node’ may be used uninitialized in this function [-Wmaybe-uninitialized]
         elem->set_node(0) = _mesh->node_ptr(start_node->id());
                             ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
</pre>

@tradowsk please fix.

tradowsk:
I shall get a PR out this afternoon. Weird the warnings don't show up on GCC5.3. Will test with GCC6.1 from now on

tradowsk:
I just rebuilt the stack and compiled GRINS with GCC6.1, and I did not get a warning from the rayfire.

Would the `METHOD` affect that? I'm using `dbg` right now

pbauman:
`METHOD=devel`

pbauman:
Closed by #444

PR #444, opened 2016-08-25
RayfireMesh GCC warnings fix

tradowsk:
This fixes warnings from GCC6.1 pointed out by @pbauman in #441.

For some context, on coarsening, we know the coarsened rayfire elem has to travel across the entire `parent_elem`. But since we have already calculated the intersection points for all the child elems, we can coarsen by just finding the node closest to the origin and farthest from it. So we iterate over all the child elems of our parent, and keep track of the start and end nodes, which later become the 2 nodes of the coarsened EDGE2 rayfire elem


pbauman:
Confirmed the offending warnings are gone and `make check` passes. Thanks!

PR #445, opened 2016-08-31
Integrated Function QoI

tradowsk:
The `IntegratedFunction` class builds upon `RayfireMesh` to evaluate a given function along a 1D rayfire. Integration is done with Gauss Quadrature (`QGauss`).

The template parameter can be `FunctionBase` or `FEMFunctionBase`, depending on whether or not the `context` is needed to evaluate the function. An object of the templated type is then passed in to the constructor, and will be used to evaluate the function at the given quadrature points along the rayfire.

`IntegratedFunction` was then added to the `QoIFactory`, which allows us to run everything directly from an input file. User adds a `[QoI]` section with `integrated_function`, and must specify:

1. Rayfire origin coordinates (currently restricted to 2D)
2. Rayfire polar angle, theta
3. Function to be evaluated
4. Quadrature order (default order is 2 if omitted)

`element_qoi_derivative()` is currently not implemented, as there are no readily-available hooks in the templated classes for derivative evaluation. This will be addressed in an upcoming PR.

Also, mesh refinement through the input file is not yet supported, but will be in a future PR.
However, the convergence test in `integrated_function_test` does refinement manually, so the functionality is there and works, just not yet integrated with the refinement using an input file.

The first 2 commits address Valgrind complaints.

`make check` passes on GCC 6.1 in dbg, devel, and opt

### DIFF
--- a/src/qoi/include/grins/integrated_function.h
+++ b/src/qoi/include/grins/integrated_function.h
@@ -1,0 +1,123 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// GRINS - General Reacting Incompressible Navier-Stokes
+//
+// Copyright (C) 2014-2016 Paul T. Bauman, Roy H. Stogner
+// Copyright (C) 2010-2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+
+#ifndef GRINS_INTEGRATED_FUNCTION_H
+#define GRINS_INTEGRATED_FUNCTION_H
+
+// libMesh
+#include "libmesh/quadrature.h"
+#include "libmesh/exact_solution.h"
+#include "libmesh/fem_function_base.h"
+
+// GRINS
+#include "grins/qoi_base.h"
+#include "grins/variable_name_defaults.h"
+#include "grins/rayfire_mesh.h"
+
+namespace GRINS
+{
+
+  //! IntegratedFunction
+  /*!
+  The purpose of this class is to integrate a function along a 1D line through a 2D mesh.
+  This class utilizes the RayfireMesh class to calculate the line and provide 1D elements for evaluating
+  the function.
+
+  The template parameter must be FunctionBase or FEMFunctionBase.
+
+  Currently, calculating element qoi derivatives is not supported.
+  */
+  template<typename Function>
+  class IntegratedFunction : public QoIBase
+  {
+  public:
+    //! Constructor
+    /*!
+    @param p_level The desired Gauss Quadrature level
+    @param f A FunctionBase or FEMFunctionBase object for evaluting the QoI
+    @param rayfire A RayfireMesh object (will be initialized in init())
+    @param qoi_name Passed to the QoIBase
+    */
+    IntegratedFunction(unsigned int p_level, SharedPtr<Function> f, SharedPtr<RayfireMesh> rayfire, const std::string& qoi_name);
+
+    //! Required to provide clone (deep-copy) for adding QoI object to libMesh objects.
+    virtual QoIBase* clone() const;
+
+    virtual bool assemble_on_interior() const;
+
+    virtual bool assemble_on_sides() const;
+
+    //! Compute the qoi value.
+    virtual void element_qoi( AssemblyContext& context,
+                              const unsigned int qoi_index );
+
+    //! Compute the qoi derivative with respect to the solution.
+    /*!
+    Currently not implemented
+    */
+    virtual void element_qoi_derivative( AssemblyContext& context,
+                                         const unsigned int qoi_index );
+    //! Initializes the rayfire with the mesh from system
+    virtual void init( const GetPot& input,
+                       const MultiphysicsSystem& system,
+                       unsigned int qoi_num );
+
+
+  private:
+    //! Quadrature order
+    unsigned int _p_level;
+
+    //! Pointer to the template class used for function evaluation
+    SharedPtr<Function> _f;
+
+    //! Pointer to RayfireMesh object
+    SharedPtr<RayfireMesh> _rayfire;
+
+    //! QBase object for adding quadrature to rayfire elements
+    SharedPtr<libMesh::QBase> _qbase;
+
+    //! Compute the value of a QoI at a QP
+    libMesh::Real qoi_value(Function& f,AssemblyContext& context,const libMesh::Point& xyz);
+
+    //! User cannot call empty constructor
+    IntegratedFunction();
+
+  };
+
+  template<typename Function>
+  inline
+  bool IntegratedFunction<Function>::assemble_on_interior() const
+  {
+    return true;
+  }
+
+  template<typename Function>
+  inline
+  bool IntegratedFunction<Function>::assemble_on_sides() const
+  {
+    return false;
+  }
+}
+#endif //GRINS_INTEGRATED_FUNCTION_H
--- a/src/qoi/src/integrated_function.C
+++ b/src/qoi/src/integrated_function.C
@@ -1,0 +1,133 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// GRINS - General Reacting Incompressible Navier-Stokes
+//
+// Copyright (C) 2014-2016 Paul T. Bauman, Roy H. Stogner
+// Copyright (C) 2010-2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+
+// This class
+#include "grins/integrated_function.h"
+
+// GRINS
+#include "grins/multiphysics_sys.h"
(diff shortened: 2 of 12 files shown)


pbauman:
Sorry to be a pain, but can you cherry pick  	60897d3 and  	a57efce to a separate PR and then we'll rebase this after that one is merged? They're logically separate from this.

src/qoi/include/grins/integrated_function.h
```diff
@@ -0,0 +1,125 @@
...
+  /*!
+  The purpose of this class is to integrate a function along a 1D line through a 2D mesh.
+  This class utilizes the RayfireMesh class to calculate the line and provide 1D elements for evaluating
+  the function.
+
+  The template parameter must be FunctionBase or FEMFunctionBase.
+
+  Currently, calculating element qoi derivatives is not supported.
+  */
+  template<typename func>
```
pbauman: Can you `sed` and change `func` to `Function`?

src/qoi/src/integrated_function.C
```diff
@@ -0,0 +1,140 @@
...
+{
+  template<typename func>
+  IntegratedFunction<func>::IntegratedFunction(unsigned int p_level, SharedPtr<func> f, SharedPtr<RayfireMesh> rayfire, const std::string& qoi_name) :
+    QoIBase(qoi_name),
+    _p_level(p_level),
+    _f(f),
+    _rayfire(rayfire)
+  {
+    // use Gauss Quadrature
+    _qbase = new libMesh::QGauss(1,libMesh::Order(_p_level));
```
pbauman: I'd do `_qbase.reset(new libMesh::QGauss(1,libMesh::Order(_p_level));` to make it a bit more clear this is a smart pointer.

src/qoi/src/integrated_function.C
```diff
@@ -0,0 +1,140 @@
...
+  template<typename func>
+  void IntegratedFunction<func>::init
+    (const GetPot& /*input*/,
+     const MultiphysicsSystem& system,
+     unsigned int /*qoi_num*/ )
+  {
+    _rayfire->init(system.get_mesh());
+  }
+
+  template<typename func>
```
pbauman: This isn't a pure virtual function so if we're not overriding it with anything, then don't even bother declaring it.

src/qoi/src/integrated_function.C
```diff
@@ -0,0 +1,140 @@
...
+  void IntegratedFunction<func>::element_qoi_derivative( AssemblyContext& /*context*/,
+                                       const unsigned int /*qoi_index*/ )
+  {
+    //TODO
+    libmesh_not_implemented();
+  }
+
+
+  // private
```
pbauman: Unnecessary.

src/qoi/src/integrated_function.C
```diff
@@ -0,0 +1,140 @@
...
+
+  // speciaizations of the qoi_value() function
+  template<>
+  libMesh::Real IntegratedFunction<libMesh::FEMFunctionBase<libMesh::Real>>::qoi_value(libMesh::FEMFunctionBase<libMesh::Real>& f,AssemblyContext& context,const libMesh::Point& xyz)
```
pbauman: We're not officially C++11 only yet, so you need a space between the two `>>`, i.e. `IntegratedFunction<libMesh::FEMFunctionBase<libMesh::Real>>` should be `IntegratedFunction<libMesh::FEMFunctionBase<libMesh::Real> >`

src/qoi/src/integrated_function.C
```diff
@@ -0,0 +1,140 @@
...
+  {
+    return f(context,xyz);
+  }
+
+  template<>
+  libMesh::Real IntegratedFunction<libMesh::FunctionBase<libMesh::Real>>::qoi_value(libMesh::FunctionBase<libMesh::Real>& f,AssemblyContext& /*context*/,const libMesh::Point& xyz)
```
pbauman: Also here.

test/unit/integrated_function_test.C
```diff
@@ -0,0 +1,274 @@
...
+      theta.push_back(0.0);
+      theta.push_back(0.25);
+
+      GRINS::SimulationBuilder sim_builder;
+
+      GRINS::Simulation sim(input,
+                            empty_command_line,
+                            sim_builder,
+                            *TestCommWorld );
+
```
pbauman: Can you make a helper function to reduce code duplication of these beginning blocks?

test/unit/integrated_function_test.C
```diff
@@ -0,0 +1,274 @@
...
+                      if (err < libMesh::TOLERANCE)
+                        {
+                          num_converged++;
+                          converged[i] = true;
+                        }
+                      else if (errors[i] == -1.0)
+                        errors[i] = err;
+                      else if (errors[i] > -1.0)
+                        {
+                          // check for linear convergence
```
pbauman: I don't understand the second assert. What exactly are you testing here?

test/unit/integrated_function_test.C
pbauman: Actually, you can probably just override `setUp()` (I think that's the name) and that will be called before each test. Similarly for `tearDown()` (where you could put the `VariableWarehouse` clearing.

src/qoi/src/qoi_factory.C
```diff
@@ -108,6 +112,36 @@ namespace GRINS
...
+        unsigned int rayfire_dim = input.vector_variable_size("QoI/IntegratedFunction/Rayfire/origin");
+
+        if (rayfire_dim != 2)
+          libmesh_error_msg("ERROR: Only 2D Rayfires are currently supported");
+
+        libMesh::Point origin;
+        origin(0) = input("QoI/IntegratedFunction/Rayfire/origin", 0.0, 0);
+        origin(1) = input("QoI/IntegratedFunction/Rayfire/origin", 0.0, 1);
+
+        libMesh::Real theta = input("QoI/IntegratedFunction/Rayfire/theta", -7.0);
```
pbauman: I'd rather a check `input.have_variable(...)` and error out if it's not found.

src/qoi/src/qoi_factory.C
```diff
@@ -108,6 +112,36 @@ namespace GRINS
...
+        if (theta == -7.0)
+          libmesh_error_msg("ERROR: Invalid/undefined spherical polar angle theta for Rayfire");
+
+        if (input.have_variable("QoI/IntegratedFunction/Rayfire/phi"))
+          libmesh_error_msg("ERROR: cannot specify spherical azimuthal angle phi for Rayfire, only 2D is currently supported");
+
+        std::string function = input("QoI/IntegratedFunction/function", "");
```
pbauman: Same here.

pbauman:
OK, thanks. Only other thing is could you open an issue for adding a `IntegratedFEMFunction` QoI and reference this PR? It would be easy to add, and we'd obviously want to test it, but it can be separate from this PR.

test/unit/integrated_function_test.C
tradowsk: I am using this definition of linear convergence:

\mu = \frac{ |x_{n+1} - x_{n}| }{ |x_{n} - x_{n-1}| }

where 0 < \mu < 1 for convergence on each step.

The first assert I guess is more of a sanity check, since numerator and denomerator of \mu are abs(). The second assert makes sure the error of the current step is less than the error of the previous step, or \mu < 1

pbauman: OK. Wrong convergence measure. What you have there would be more appropriate for things like iterative solvers. The error here should be |e| =  |u^h - u| \le C h, where C is some unknown constant. So, if you look at log e vs. log h, you should have a line with slope 1, i.e. linear convergence.

tradowsk: Sorry, for reference, I was using the following convergence metric, not the one in my previous note:

\mu = \frac{ |x_{n+1} - x*| }{ |x_{n} - x*| }

where x* is the known analytical solution

tradowsk:
Alright, I think I added all the fixes that @pbauman requested, and I rebased on master.

pbauman:
Looks like fb3ff74 comes after you actually use it in d4ddc2e

tradowsk:
> Looks like fb3ff74 comes after you actually use it in d4ddc2e

That's odd, it showed up in the right order on my `git log`, and `git push --force` said everything was up to date.

I rebased again to get new hashes, and it looks like everything is fine now.

test/unit/integrated_function_test.C
```diff
@@ -0,0 +1,309 @@
...
+
+    CPPUNIT_TEST( test_exact_answer );
+    CPPUNIT_TEST( test_convergence );
+
+    CPPUNIT_TEST_SUITE_END();
+
+  public:
+
+    void setUp()
+      {
```
pbauman: This supposed to be empty?

test/unit/integrated_function_test.C
```diff
@@ -0,0 +1,309 @@
...
+                }
+
+              CPPUNIT_ASSERT(h > 0.0);
+
+              for (unsigned int i=0; i<calc_answers.size(); i++)
+                {
+                  if (!converged[i])
+                    {
+                      libMesh::Real err = std::abs( _sim->get_qoi_value(i) - calc_answers[i] );
+                      if (err < 2.5e-9)
```
pbauman: I would make this an input parameter to the function. Then, you can control it for the different tests. Weren't you saying one of them takes longer than the other (sin vs. exp or something)? Also, magic numbers bad.

test/unit/integrated_function_test.C
```diff
@@ -0,0 +1,309 @@
...
+                  rayfire->reinit(system->get_mesh());
+                  es->reinit();
+                }
+
+            } while(num_converged < calc_answers.size());
+
+            for (unsigned int i=0; i<functions.size(); i++)
+              {
+                libMesh::Real convergence_rate = this->linear_slope(h_vals[i],errors[i],0,errors[i].size()-1);
+                CPPUNIT_ASSERT_DOUBLES_EQUAL(4.0,convergence_rate,0.08);
```
pbauman: Can we abstract this out some how into the `RegressionHelper` to have a test for the rate?

test/unit/integrated_function_test.C
pbauman: Perhaps you meant to put some of the comment setup in here?

test/unit/integrated_function_test.C
tradowsk: They both take about the same number of iterations now. I had to pick the tolerance carefully so that the exponential didn't need another refinement, since that more than doubled the runtime. I can have a separate tolerance for each function if you want

test/unit/integrated_function_test.C
tradowsk: I took the stuff from here and moved it to `init_sim()` because I needed the `filename` parameter since the QoI test uses a different input file. Squashed in deleting it

test/unit/integrated_function_test.C
tradowsk: You mean pass the raw data to the RegressionHelper along with a tolerance and rate and have it do the CPPUNIT_ASSERT?

pbauman: Sure. Something that will be convenient to reuse.

test/unit/integrated_function_test.C
tradowsk: I added a `tolerance` vector and set it for each function, so the tol values are nice and clean now.

pbauman:
OK, this ready to go?

tradowsk:
Fails in `opt` mode... will investigate with valgrind

tradowsk:
#454 fixes the issue that caused the aforementioned failure

tradowsk:
Rebased on master after #454. All tests passing

PR #446, opened 2016-09-07
Rayfire memleak fix

tradowsk:
These commits address memory leak complaints from Valgrind.

Fixes include removing unnecessary pointers in the `RayfireMesh` class and fixing an `invalid read` in the `rayfireAMR_test`. Neither of these affected the operation of the `RayfireMesh`, just the performance.


pbauman:
Thanks!

PR #448, opened 2016-09-08
Skip antioch test when configured without Antioch

roystgnr:


pbauman:
Sorry for that. Thanks for the fix.

PR #449, opened 2016-09-09
Rayfire start_elem fix

tradowsk:
The `libMesh::PointLocatorBase` is used to find the elem of the mesh that contains the `origin` point of the rayfire.

However, the `PLB` returns the first elem it finds that contains the given `origin` point. When the origin is on an elem boundary, then multiple elems contain that point. Since PLB only returns the first elem it finds, it may not be the correct `start_elem` for the rayfire.

An additional check is added to find the correct starting elem. We walk a short distance along the rayfire, and check if we are still within the elem returned from the `PLB`. If not, then we do the same check with all neighbors of that elem to determine if any of them are along the rayfire. Since the origin must be on a mesh boundary, only the elem from the `PLB` and its neighbors would be valid possibilities for being the `start_elem`.

The added `CPPUnit` test demonstrates the issue, and verifies that the bugfix works correctly.

`make check` passes with `GCC 6.1` with `dbg` & `devel`

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -79,8 +79,7 @@ namespace GRINS
     libMesh::Point start_point(_origin);
 
     // get first element
-    libMesh::UniquePtr<libMesh::PointLocatorBase> locator = mesh_base.sub_point_locator();
-    const libMesh::Elem* start_elem = (*locator)(_origin);
+    const libMesh::Elem* start_elem = this->get_start_elem(mesh_base);
 
     if (!start_elem)
       libmesh_error_msg("Origin is not on mesh");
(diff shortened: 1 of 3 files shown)


src/qoi/src/rayfire_mesh.C
```diff
@@ -199,6 +198,35 @@ namespace GRINS
...
+              continue;
+
+            if (this->rayfire_in_elem(_origin,neighbor_elem))
+              {
+                start_elem = neighbor_elem;
+                break;
+              }
+          }
+      }
+
```
pbauman: Add an assert here that `start_elm` is now valid.

pbauman:
Squash in the assert please, then I'll merge. Thanks!

src/qoi/src/rayfire_mesh.C
tradowsk: I check the return value of `get_start_elem()` in `init()` line 85, because it can return `NULL` if the origin is not on the mesh, in which case, I `libmesh_error_msg`

tradowsk: Don't merge yet, I have to squash in code to handle when `elem` from the `PLB` is null.

tradowsk:
Squashed in small fix for when `PLB` returns `NULL`, indicating origin is not even on the mesh.

src/qoi/src/rayfire_mesh.C
pbauman: Hmm, OK. In that case, please add a comment both at the return line and the function documentation indicating that this may return a NULL element.

tradowsk: Squashed in

tradowsk:
I made another change to this branch that apparently didn't get squashed in. Literally one line of [code](https://github.com/tradowsk/grins/blob/rayfire-pointlocator-fix/src/qoi/src/rayfire_mesh.C#L350) just needs to be deleted. That function can also return NULL if the rayfire exits on a node that is not a corner, meaning it needs to make it to that final `return` statement as it will never hit the return on line 340. Do you want a separate PR for that?

pbauman:
I'll just push it to master.

tradowsk:
Thanks, sorry about the mixup

pbauman:
Done.

PR #452, opened 2016-09-14
If we've got MUMPS, let's use it here

roystgnr:
I can't find any other way to get proper convergence in parallel.
This should enable LIBMESH_RUN='mpifoo -bar' for MUMPS users but
without breaking serial tests for MUMPS-free PETSc builds.


roystgnr:
#453 is much better.

PR #453, opened 2016-09-14
Update suspended_cable_test options

### LINKED ISSUE #447: suspended_cable_test.sh fails in parallel
It breaks even on 2 processors for me.  I can't seem to get it to converge reliably in parallel without MUMPS LU.

pbauman:
Cholesky sub_pc is not robust in parallel as the PC dies due
to indefiniteness using pc bjacobi. So, be gross and just use
GMRES with LU. This runs on 1-16 processors for me.

Fixes #447 and I think replaces #452. @roystgnr please confirm and sorry I didn't do this sooner.


roystgnr:
Huh?  I could have sworn I tried with GMRES+LU and it didn't work.  But now this works for me, for 1 through 24 processors.  Much better than my solution.

pbauman:
Thanks.

PR #454, opened 2016-09-14
Rayfire origin point

tradowsk:
The RayfireMesh constructor used to take a `libMesh::Point& origin`. But, if the origin point passed in goes out of scope, as happens when creating the `IntegratedFunction` class in `qoi_factory` in #445 	486c060, then `rayfire._origin` becomes undefined.

Instead of just setting `_origin` to that reference point, we copy it in the initialization list so that `rayfire._origin` cannot be changed outside the class.

`make check` still passes in `dbg, devel, opt`, and this fixes the failure mentioned in #445


src/qoi/src/rayfire_mesh.C
```diff
@@ -40,7 +40,7 @@ namespace GRINS
 {
   RayfireMesh::RayfireMesh(libMesh::Point& origin, libMesh::Real theta, libMesh::Real phi) :
     _dim(3),
-    _origin(origin),
+    _origin(libMesh::Point(origin)),
```
pbauman: Huh? You should be able to just pass origin and then `_origin` will be created from the copy constructor of `origin`. That needlessly introduces another copy construction.

src/qoi/src/rayfire_mesh.C
```diff
@@ -50,7 +50,7 @@ namespace GRINS
 
   RayfireMesh::RayfireMesh(libMesh::Point& origin, libMesh::Real theta) :
     _dim(2),
-    _origin(origin),
+    _origin(libMesh::Point(origin)),
```
pbauman: Here too.

src/qoi/src/rayfire_mesh.C
tradowsk: Okay, didn't realize I was copying twice. Squashed in the fix, and `make check` still passes.

PR #455, opened 2016-09-14
Never pass negative temperature to Antioch kinetics

roystgnr:
As we can see from the fact that this changed the ozone flame
regression results, this is a real problem in combustion problems.

I'm planning on making a similar change for mass/molar fractions, but before I do I thought I'd run this one up the flagpole and make sure I'm not going in the wrong direction.  "This should be done on the Antioch level" or "you need to do _foo_ to the stabilization parameters to avoid getting in this mess", perhaps?

This doesn't actually clip the temperature that GRINS stores, just the temperature passed to Antioch kinetics, so it shouldn't screw with energy conservation.


pbauman:
Da fuq? I'm surprised that this changed the regression; we must have a tighter tolerance here than I remember. How different was the solution?

I've only ever hit negative temperatures in the early Newton steps (with pretty big time steps) or with grids that are really coarse so you get lots of over/undershoot around the flame front and for the latter, the solution died very quickly anyway.

Not opposed to the clipping, but I wonder if we should be doing it up stream?

The mass/mole fractions are different problem. We definitely get negative fractions of trace species near the flame front on coarse grids, but usually numerically small enough that it doesn't destroy convergence and one can then refine the grid to smooth everything out.

roystgnr:
No tolerance issues here - if the temperature in the solution never dips below 0K then we should get 0 diff, if not then we shouldn't.

My only worries about clipping upstream are that an additional temporary variable is more expensive on the Antioch side (where it could be a vector, could force intermediate evaluation of an expression template, etc) than on the GRINS side (where we know it's a double), and that although it's necessary for some usage cases it may just be a redundant branch instruction for others.  I don't think it's unreasonable for a chemistry library to give garbage results if you give it negative temperatures.

pbauman:
More later, but I think something was FUBARed with the existing xdr (not all sure how that happened) - test is failing in existing master because it's failing to read the xdr file. Will follow up later. Building latest libMesh to be sure.

pbauman:
Also, @permcody, ping on Civet being open-sourced. :)

permcody:
Hey @pbauman - We've escalated this up in our company. I'm very disappointed with our legal team as I understand they are quite literally letting the approval for this product sit on their desks most likely due to fear, uncertainty, trust issues. I had a meeting with Rich and our export control folks to get additional people in the loop. I plan on following up with our new OSS expert we've hired into the company. Again, I apologize for this continued failure to get this product out the door. I really am trying and you'll be the first to know when it's available.

pbauman:
Thanks for the update @permcody. I don't mean to be a (ugly) pest - just anxious to get it going and hooked in here (and in libMesh, Antioch and other places). BuildBot looks pretty cumbersome to get to work with GitHub and Jenkins looked pretty restrictive, so I'm just anxious to try out Civet.

pbauman:
OK, so definitely failing on master with latest libMesh and looks like old vs. new libmesh xdr compatibility. I will open up a PR with the temporary fix. I have a meeting in a few mins, but after that, I'm going to work on cleaning up the regression test suite. Right now what's being done is stupid - we're not reading the same way that grins writes so separate xdr has to be generated instead of what's written out by grins. I will fix that and update all the shit load of gold data so it's fixed once and for all. Then, we can revisit the clipping.

permcody:
Paul, I don't blame you for continuing to ping me for status. I'm actually using that fact with individuals at the lab to apply more pressure. I've been telling them that I have an external collaborator who's purchased hardware in anticipation of using this product. I think that helps, but we're stepping up the pressure even more.

roystgnr:
I'll edit this to get rid of the xdr change and make sure it still works.

roystgnr:
Looks like clipping temperatures doesn't change the regression results, but clipping densities does.  I'll update this PR shortly.

roystgnr:
This passes "make check" for me, but that must mean I'm not understanding #456, because IIRC all I did this time was the same thing I did in #450: cp the output from the test run to be the new gold standard file.

roystgnr:
@pbauman, would you take another look at this?  I made all the behaviors optional and non-default, so hopefully there shouldn't be any downside to merging.

pbauman:
Damn! I had to take down femputer before I left for the undergrads exam and haven't had a chance to restart it. Otherwise, you would've gotten some grins-devel testing! I'll take a look now.

pbauman:
This looks good. Only nitpick would be to add documentation of the new Antioch options to the master example input file.

Apparently I can't VPN using my AT&T hot spot at the airport, so I can't restart femputer until later tonight. Go ahead and merge one you push the master example input file documentation. Thanks!

roystgnr:
Damn, that's not a nitpick, that's something I swear I intended to do but then didn't get to.  Will do now.

PR #456, opened 2016-09-19
Update ozone xdr data

pbauman:
It appears that the ozone xdr data become incompatible with newer
versions of libMesh. I checked the solution and it was as expected.

Mentioned in #455. This is temporary fix while I rework how we handle regression data (and before pushing forward with #455 to confirm that there acutally was a change in the regression solution). Note that I was seeing the failure in the current master of GRINS using the current master of libMesh. Likely this incompatibility crept in and we didn't have automated coverage to check it. See discussion with in #455.


roystgnr:
> It appears that the ozone xdr data become incompatible with newer
versions of libMesh.

This sounds like a bug we ought to be tracking down, not working around.  Any idea *what* the incompatibility was?

pbauman:
OK, I think I see what happened and it's 99% my fault. Confirming now.

roystgnr:
I can replicate the failure with master+master, but I've no idea what could have happened.

In the future, though, we might want to stick with xda.gz rather than xdr; it's just about as efficient and it's much easier to debug.

pbauman:
It's related to the fact that I made the regression test driver not read with the same options as what we write with GRINS proper. So, in #450, I think the gold data you generated was just a copy of the normal GRINS output which is not compatible with the read in generic_regression_testing_app. I'm confirming this now. If so, this should be fine to merge and I'll clean up that bullshit in the regression tests (which is my fault); could take a bit - quite a few tests to clean up.

pbauman:
So 99% my fault. The other 1% is INL for not having Civet released yet. :P

pbauman:
Confirmed. All bueno with libMesh master at GRINS 8c50022 and failure with libMesh master at GRINS 8c50022. I think this is safe to merge and I'll work on making the regression data handling less stupid.

pbauman:
Actually, no. What I can do instead is set up the framework in a different PR and have the first test moved to that framework be this ozone one and just ditch this PR.

roystgnr:
Can we just pull this one?

pbauman:
OK.

roystgnr:
I've got #455 and its successor waiting in the wings.

PR #457, opened 2016-09-26
Update QoI Output and Corresponding Options

pbauman:
Long overdue, this PR adds the capability to output the QoI(s) to a text file as well as the terminal. The `QoIOutput` class handles parsing the options and deciding what modes to output - the actual QoI printing is still delegated to the QoI.

This will be used in refactoring the regression tests so that, for those tests with a QoI, we can just read in the QoI text files output from the grins run and compare to the gold values. Thus, right now, only the `print_qoi` option is actually tested in the test suite, but I've manually verified the file output in the `sa_airfoil` example.

Implicit in here is starting down the resectioning of output options in the input; hence the new `OutputParsing` class. Right now only the `[Output/QoI]` parts are there, but we might as well discuss this now so that I can change the QoI parts if needed. Here is what I was thinking.

<pre>
[Output]
   [./Display]
      ...all options relevant to terminal output...
   [../]
   [./QoI]
       #Outputs qoi values to file_prefix.dat txt file. If no file_prefix value is specified, QoIs not output to file.
       file_prefix = 'myqoi'
   [../]
   [./Visualization]
      file_prefix = ''
      formats = ''
      output_residual = ''
      output_adjoint = ''
      ...etc...
   [../]
   [./Mesh]
      file_prefix = ''
      formats = ''
   [../]
   [./Restart]
      file_prefix = ''
      formats = ''
   [../]
[]
</pre>

Thoughts @roystgnr?

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -199,6 +199,7 @@ libgrins_la_SOURCES += qoi/src/parsed_interior_qoi.C
 libgrins_la_SOURCES += qoi/src/weighted_flux_qoi.C
 libgrins_la_SOURCES += qoi/src/rayfire_mesh.C
 libgrins_la_SOURCES += qoi/src/integrated_function.C
+libgrins_la_SOURCES += qoi/src/qoi_output.C
 
 # src/solver files
 libgrins_la_SOURCES += solver/src/grins_solver.C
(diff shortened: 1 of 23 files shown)


roystgnr:
Very convenient, thanks!

In my own codes I've found it handy to be able to output multiple QoIs to independent files; we might want to keep that in mind for future addition here.  Maybe 'file_prefix' could be 'default_file_prefix' and we could later override or append on a per-qoi basis?

pbauman:
> Maybe 'file_prefix' could be 'default_file_prefix' and we could later override or append on a per-qoi basis?

Excellent suggestion. I'll make that change, thanks.

pbauman:
OK, tacked on a commit to use `default_file_prefix` instead.

roystgnr:
:+1: when you're ready to merge.

nicholasmalaya:
Late to this party (for obvious reasons) but this is awesome, thanks for doing this. I'll be playing around with this in the future and will comment if I notice anything. :+1:

PR #458, opened 2016-09-29
1-D fixes for Reacting Low Mach Navier-Stokes

roystgnr:
Not yet sure that this is complete, and I'll probably end up saying "forget it, let's just make a strip of quads" before it is, but I think this is at least a good start.


pbauman:
Sorry for the delay @roystgnr. These were more innocuous than I was anticipating. Thanks.

roystgnr:
No problem.  This actually *isn't* a complete set of fixes; otherwise I'd have been agitating to have it merged.  This gets us running but doesn't get us correct answers.

PR #466, opened 2016-11-15
VariablePinning physics

roystgnr:
I didn't end up using this, but it works as advertised and others might find it useful.


pbauman:
Damn. Maybe this is the way we should do the pressure pinning, i.e. instead of an option in the Navier-Stokes physics class.

pbauman:
Thanks @roystgnr. I think this will be useful, e.g. to make pressure pinning more clear.

PR #468, opened 2016-12-05
Ignore third party library code in compiler warnings

roystgnr:
GRVY and CPPUnit use auto_ptr, and probably will for the foreseeable future.  But with the thousands of lines of "auto_ptr!  I saw an auto_ptr!  Did you know that's deprecated!?" disabled, I can actually read GCC warnings about our *own* code again.


pbauman:
Sorry for the delay on this @roystgnr. Thanks.

PR #473, opened 2016-12-14
Parsed qoi regression test update

roystgnr:
Yet again I caught us using 1e-10 solution tolerances for a solve that was getting done with the higher default solver tolerances.  For many LIBMESH_RUN settings I was getting test failures with 1e-9 or so error.

This time I can't figure out how I didn't run into the problem earlier, but dropping the solver tolerances and regenerating the solution fixes it.


roystgnr:
@pbauman I know I've got a lousy track record for updating gold standard files this week, but I think this one is good to merge.

pbauman:
I'm mostly to blame because of the crappy initial configuration of how those worked (and my failure to update the rest to similar format). Hopefully I'll find a groove here soon.

PR #476, opened 2017-01-25
Skip SpectrscopicAbsorption test if not configured with Antioch

tradowsk:
These tests require Antioch to calculate the QoI, so if GRINS was not configured with Antioch, these tests need to be skipped.

### DIFF
--- a/test/unit/spectroscopic_absorption_test.C
+++ b/test/unit/spectroscopic_absorption_test.C
@@ -53,10 +53,10 @@ namespace GRINSTesting
   {
   public:
     CPPUNIT_TEST_SUITE( SpectroscopicAbsorptionTest );
-
+#if GRINS_HAVE_ANTIOCH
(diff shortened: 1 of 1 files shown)


pbauman:
OK. I haven't setup the unit-testing-per-optional-package structure yet so we can see what unit tests are skipped, so I won't hold you to that. I haven't setup all the CIVET variants yet (of which no optional packages was one) - did you compile this and run `make check` in such a configuration?

tradowsk:
Yes, and I get a `unit_driver` failure for both these tests with this message: `..Error: Antioch not enabled in this configuration. Reconfigure using --with-antioch option...`

pbauman:
Sorry, I was driving at that your changes now get `make check` passing in this configuration?

tradowsk:
Oh sorry. Yes, with this fix `make check` passes successfully

pbauman:
Cool, thanks.

PR #477, opened 2017-01-30
RayfireMesh fixes to handle special cases

tradowsk:
Recent work uncovered 2 bugs in the RayfireMesh class with non-square QUAD elements.

The first fix is for the case where the rayfire travels the boundary between 2 elements, one of which is to be refined. When updating the rayfire to reflect the refinement, it was possible for `RayfireMesh::refine()` to incorrectly identify elements along the rayfire due to the rayfire being along an element boundary. This caused the rayfire to wander into other elements, creating a hole in the rayfire that ultimately led to null pointer exceptions. This is fixed by adding a flag to certain private class functions to enforce when elements must have the same parent, as is the case when the rayfire is refined.

The second issue arises when the rayfire passes very close to vertices or element boundaries (within `libMesh::TOLERANCE`) after refinement. In certain instances, the correct `next_elem` would not be identified due to the default tolerances used in various functions. These issues are fixed by tightening the tolerance on `Elem::contains_point()` and `RayfireMesh::rayfire_in_elem()`. Two test cases are added to identify these issues, with the second one (`refine_deformed_elem_near_tolerance()`) coming from a larger mesh that originally demonstrated the issue. The coordinates given were obtained from the DDT debugger, and attempts to remove digits resulted in failures in `libMesh::Elem::contains_point()` and `PointLocatorBase::operator()`, used in verifying the validity of the supplied origin point.

`make check` passes in `dbg`,`devel`, & `opt` with `GCC 6.1.0`. Fixes were also verified on the original mesh that uncovered these issues, and 6 AMR steps were performed with no issues. Originally, the run failed at AMR step 3.

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -58,7 +58,6 @@ namespace GRINS
       libmesh_error_msg("Please supply a theta value between -2*pi and 2*pi");
   }
 
-
   void RayfireMesh::init(const libMesh::MeshBase& mesh_base)
   {
     // consistency check
@@ -74,8 +73,6 @@ namespace GRINS
 
     _mesh = new libMesh::Mesh(mesh_base.comm(),(unsigned char)1);
 
-    unsigned int node_id = 0;
-
     libMesh::Point start_point(_origin);
 
     // get first element
@@ -89,7 +86,8 @@ namespace GRINS
     this->check_origin_on_boundary(start_elem);
 
     // add the origin point to the point list
-    _mesh->add_point(start_point,node_id++);
+    libMesh::Node * start_node = _mesh->add_point(start_point);
+    libMesh::Node * end_node = NULL;
 
     libMesh::Point end_point;
 
@@ -103,15 +101,16 @@ namespace GRINS
         next_elem = this->get_next_elem(prev_elem,start_point,end_point);
 
         // add end point as node on the rayfire mesh
-        _mesh->add_point(end_point,node_id);
+        end_node = _mesh->add_point(end_point);
         libMesh::Elem* elem = _mesh->add_elem(new libMesh::Edge2);
-        elem->set_node(0) = _mesh->node_ptr(node_id-1);
-        elem->set_node(1) = _mesh->node_ptr(node_id++);
+        elem->set_node(0) = start_node;
+        elem->set_node(1) = end_node;
 
         // add new rayfire elem to the map
         _elem_id_map[prev_elem->id()] = elem;
 
         start_point = end_point;
+        start_node = end_node;
         prev_elem = next_elem;
       } while(next_elem);
 
@@ -178,6 +177,7 @@ namespace GRINS
     // coarsen the elements that need it
     for (unsigned int i=0; i<elems_to_coarsen.size(); i++)
       this->coarsen(elems_to_coarsen[i]);
+
   }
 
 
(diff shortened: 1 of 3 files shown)


src/qoi/src/rayfire_mesh.C
```diff
@@ -103,15 +101,27 @@ namespace GRINS
...
-        elem->set_node(0) = _mesh->node_ptr(node_id-1);
-        elem->set_node(1) = _mesh->node_ptr(node_id++);
+        elem->set_node(0) = start_node;
+        elem->set_node(1) = end_node;
+        
+        // warn if rayfire elem is shorter than TOLERANCE
+        if ( (start_point-end_point).norm() < libMesh::TOLERANCE)
+          {
+            libmesh_do_once(
+              std::cout <<"********\n"
```
pbauman: Instead of using `libmesh_do_once()`, you can use `libmesh_warning()` which can take a string for your message, but will also print out line, file, etc. and internally uses `libmesh_do_once`.

src/qoi/src/rayfire_mesh.C
```diff
@@ -257,23 +270,25 @@ namespace GRINS
...
+    libmesh_assert(cur_elem);
+
     libMesh::Point intersection_point;
 
     // loop over all sides of the elem and check each one for intersection
     for (unsigned int s=0; s<cur_elem->n_sides(); s++)
       {
         const libMesh::UniquePtr<libMesh::Elem> edge_elem = cur_elem->build_edge(s);
-        if (edge_elem->contains_point(start_point))
+        if (edge_elem->contains_point(start_point,libMesh::TOLERANCE*0.1))
```
pbauman: Add some documentation here about the choice for the tolerance, please.

src/qoi/src/rayfire_mesh.C
```diff
@@ -306,9 +321,12 @@ namespace GRINS
     libmesh_assert(elem);
 
     // move a little bit along the rayfire
-    // and see if we are still in the elem
-    libMesh::Real L = elem->hmin();
-    L *= 0.1;
+    // and see if we are in the elem
+    libMesh::Real L = 2*libMesh::TOLERANCE;
```
pbauman: Some documentation about this choice, please.

src/qoi/src/rayfire_mesh.C
```diff
@@ -457,14 +520,23 @@ namespace GRINS
...
+                  }
+              }
 
             if (is_vertex)
               {
                 // move a little bit along the rayfire
                 // and see if we are in the elem
-                libMesh::Real L = main_elem->child(i)->hmin();
-                L *= 0.1;
+                libMesh::Real L = 2*libMesh::TOLERANCE;
```
pbauman: Documentation about tolerance choice, please,

src/qoi/src/rayfire_mesh.C
```diff
@@ -500,24 +572,33 @@ namespace GRINS
...
+
+        elem->set_node(0) = prev_node;
+        elem->set_node(1) = new_node;
+
+        libmesh_assert_less( (*(elem->get_node(0))-_origin).norm(),  (*(elem->get_node(1))-_origin).norm());
+        
+        // warn if rayfire elem is shorter than TOLERANCE
+        if ( (start_point-end_point).norm() < libMesh::TOLERANCE)
+          {
+            libmesh_do_once(
```
pbauman: `libmesh_warning`

tradowsk:
Squashed in the requested changes. `make check` passing, as is the original AMR run with the larger mesh.

pbauman:
Thanks!

PR #478, opened 2017-02-08
new input option for Simulation to skip refinement calls in libMesh::EuqationSystems::reinit

tradowsk:
Follows [libMesh PR](https://github.com/libMesh/libmesh/pull/1233) to add an input option to disable coarsen/refine calls within `libMesh::EquationSystems::reinit`. Correct refinement flags are required for proper `RayfireMesh::reinit`. New option is `Strategies/MeshAdaptivity/disable_two_step_refinement`.

Tested with SpectroscopicAbsorption AMR run.

Do not merge until libMesh PR goes through.


pbauman:
OK. New minimum libMesh has is 4b96823 (which I'd updated to behind the scenes on `femputer` so this passed).

@tradowsk In a subsequent PR, can you add something to the test suite that was failing before so that we have it captured here.

PR #479, opened 2017-02-10
QoI reinit hooks in MultiphysicsSystem

tradowsk:
Hooks into `MultiphysicsSystem::reinit()` to also call `reinit()` on an attached `CompositeQoI`.

Included CPPUNIT tests fails without this, as the rayfire would not get refined along with the main mesh since `rayfire->reinit()` is not explicitly called in the test.

Also tests `libMesh::EquationSystems::reinit` flag from #478, without which this test will also fail because `es->reinit()` will clear the refinement flags on the main mesh and the rayfire will not get refined.

### DIFF
--- a/src/qoi/include/grins/integrated_function.h
+++ b/src/qoi/include/grins/integrated_function.h
@@ -84,6 +84,8 @@ namespace GRINS
                        const MultiphysicsSystem& system,
                        unsigned int qoi_num );
 
+    //! Reinitialize the rayfire
+    virtual void reinit(MultiphysicsSystem & system);
 
   private:
     //! Quadrature order
--- a/src/qoi/include/grins/qoi_base.h
+++ b/src/qoi/include/grins/qoi_base.h
@@ -78,6 +78,9 @@ namespace GRINS
 
     virtual void init_context( AssemblyContext& /*context*/ ){}
 
+    //! Reinitialize QoI
+    virtual void reinit(MultiphysicsSystem & /*system*/) {}
+
     //! Compute the qoi value for element interiors.
     /*! Override this method if your QoI is defined on element interiors */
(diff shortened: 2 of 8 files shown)


src/qoi/include/grins/qoi_base.h
```diff
@@ -78,6 +78,9 @@ namespace GRINS
 
     virtual void init_context( AssemblyContext& /*context*/ ){}
 
+    //! Reinitialize QoI
+    virtual void reinit(MultiphysicsSystem & /*system*/) const {}
```
pbauman: Why is this a `const` method? By definition we're changing underlying things in the `QoIBase` object, so I don't think it should be `const`.

src/qoi/include/grins/integrated_function.h
```diff
@@ -84,6 +84,8 @@ namespace GRINS
                        const MultiphysicsSystem& system,
                        unsigned int qoi_num );
 
+    //! Reinitialize the rayfire
+    virtual void reinit(MultiphysicsSystem & system) const;
```
pbauman: Again, I don't think this should be `const`.

tradowsk:
`CompositeQoI::get_qoi()` returns a `const QoIBase` object, so if the reinit is non-`const` you get a compiler error.

I can make a non-`const` version of `CompositeQoI::get_qoi()` if that would be better.

pbauman:
> I can make a non-const version of CompositeQoI::get_qoi() if that would be better.

Yes, please.

tradowsk:
Squashed in the non-`const` `get_qoi()` function. `make check` still passing

pbauman:
Great, thanks. Will merge once femputer is happy.

PR #481, opened 2017-03-17
Fix memory leak in SpectroscopicAbsorption QoI

tradowsk:
This PR fixes a memory leak. The internal `_mesh` in `RayfireMesh` was not getting properly destroyed because there seemed to be 3 copies of the `IntegratedFunction` class floating around, so the destructor wasn't ever called. I wasn't able to find the root cause, but this PR fixes the issue and is probably better design-wise.

674f3ac changes the internal `_elem_id_map` to not store raw `Elem*` for rayfire elems.

91c362c new constructor that passes the GetPot object through to the `RayfireMesh` class to allow it to construct itself. This is not strictly necessary I guess; we can just pass a `RayfireMesh*` to the `IntegratedFunction` constructor, which would then take control of the object. But I added this so that the rayfire would not have to be created outside `IntegratedFunction` and leave a dead pointer hanging around. 

d8c536a we use `UniquePtr`'s internally instead of `SharedPtr`'s, and add proper copy-constructors to deep-copy objects on `clone()`. We also don't need to have the `QGauss` object as a class variable, we can just create it within `element_qoi()` each time.

The final 3 commits update the IntegratedFunction test to work with the new API.

### DIFF
--- a/src/qoi/include/grins/integrated_function.h
+++ b/src/qoi/include/grins/integrated_function.h
@@ -55,18 +55,24 @@ namespace GRINS
   public:
     //! Constructor
     /*!
-      @param p_level The desired Gauss Quadrature level
-      @param f A FunctionBase or FEMFunctionBase object for evaluting the QoI
-      @param rayfire A RayfireMesh object (will be initialized in init())
-      @param qoi_name Passed to the QoIBase
+    @param p_level The desired Gauss Quadrature level
+    @param f A FunctionBase or FEMFunctionBase object for evaluting the QoI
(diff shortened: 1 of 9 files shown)


tradowsk:
I rebased to master to fix a downstream issue on another branch.

tradowsk:
Rebased on #502

pbauman:
Rebase one more time for me on current master since I just merged #502? Will have one last look at these diffs before merging.

pbauman:
OK. I reserve the right to remove the GetPot-based constructor in the future.

PR #482, opened 2017-04-11
Add Instantiations for Antioch::IdealGasMicroThermo/NASA9

pbauman:
This adds in support for using `Antioch::IdealGasMicroThermo`, with NASA9, for reacting Low Mach physics kernels. Instead of trying to compute vibrational and electronic energies directly, this uses the NASA9 curve fits to get the total cv, etc., and then subtracts off `cv_trans` and `cv_rot` to get `cv_vib`, which is needed for some transport models.

I've done a little bit of testing in examples and seems to work OK, but I haven't been able to encapsulate as a regression test yet. So we can wait to merge until that time.

Future PRs will add NASA7 versions of this (which will require some tweaks to input parsing).

This should not be merged before libantioch/antioch#240 as this PR depends on that.

Actually, we shouldn't merge this until the README is updated with the commit on the Antioch branch that's required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ an external chemistry library. While [Cantera](http://code.google.com/p/cantera/
 partially supported, [Antioch](https://github.com/libantioch/antioch) is fully
 supported.
 
+The current required minimum hash for using Antioch is libantioch/antioch@ad78595 (libantioch/antioch#240).
+
+
 Building GRINS 
 ================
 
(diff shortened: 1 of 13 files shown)


pbauman:
Right, these won't pass until libantioch/antioch#240 is merged and I update to that version on `femputer`.

pbauman:
OK, the Antioch PR was merged, I've updated the Antioch builds on `femputer` and I've pushed an update to the README which has the minimum required hash for Antioch now. I'd like to merge once `femputer` is happy.

PR #472, opened 2016-12-12
Testing femputer/CIVET

pbauman:
Using this PR to test `femputer`-hosted CIVET server and client automate PR testing. Do not merge.

Issue #469, opened 2016-12-07
CIVET Cases to Run


pbauman:
@roystgnr 

I'm setting up `femputer` with a CIVET server and client (server is up and running, working on client and software stack).

What do we want to have CIVET run for GRINS? I'm hoping to setup PR and nightly testing (this remains to be seen how to do this with CIVET).

Here were my initial thoughts.

PR:
   `make check`:
       dbg (with GLIBCXX_DEBUG, serial, with Cantera+Antioch)
       opt (serial, with Cantera+Antioch)
       opt (parallel, 4 processors, with Cantera+Antioch)
       dbg (with GLIBCXX_DEBUG, serial, without Cantera+Antioch)

Nightly:
  `make check`: dbg (with GLIBCXX_DEBUG, serial, with Cantera+Antioch)
                         dbg (with GLIBCXX_DEBUG, serial, without Cantera+Antioch)
                         devel (serial, with Cantera+Antioch)
                         opt (serial, with Cantera+Antioch)
                         opt (parallel, 2-16 (24?) processors ala the run parallel script we have, with Cantera+Antioch))
                         valgrind dbg (I'm not quite sure how to do this yet)

  `make run_examples` (this would need to be set up in the GRINS build system): 
                         dbg (with GLIBCXX_DEBUG, serial, with Cantera+Antioch)
                         devel (serial, with Cantera+Antioch)
                         opt (serial, with Cantera+Antioch)
                         opt (parallel, 4 processors, with Cantera+Antioch)

The question is against what libMesh? Do we take a special hash and only update when needed, e.g. when a GRINS PR uses a new feature in libMesh? Do we have CIVET pull and update libMesh before each PR and build against that? My thinking is for GRINS PRs we have a special hash of libMesh we build against and only update when needed and then nightly tests can pull the latest tip of libMesh master before starting.

Similarly for Antioch?

Also, what versions of dependencies for libMesh (PETSc, Boost, HDF5)? How many GCC versions? Other thoughts?

pbauman:
I should've mentioned - `femputer` doesn't have a graphics card, which means no OpenGL which means no VTK.

pbauman:
Nevermind, I just googled and it should be possible to build VTK without OpenGL, it will just take some massaging of the build script. Will get to that later.

roystgnr:
[...]

PR #484, opened 2017-04-24
Factory Initializer Objects

pbauman:
Ala libqueso/queso#508, this creates FactoryInitializer objects that statically instantiate the factory classes so they don't get stripped by static linking. I collected all of them into `SimulationInitializer`, which needs to be instantiated before `Simulation`. (A forthcoming PR is going to refactor the GRINS `main` to just call a function which does all the setup and run to ease extension by users.) I believe this captures all "properly" factoried objects, which is confirmed by building with `--enable-static --disable-shared` at configure time (I also confirmed we actually failed in this case before this PR).  We should add a CIVET recipe for this build configuration (#469, #472).


pbauman:
Will merge this afternoon barring any complaints.

test/exact_soln/vorticity_qoi.C
```diff
@@ -70,7 +71,9 @@ int main(int argc, char* argv[])
 
   // Initialize libMesh library.
   libMesh::LibMeshInit libmesh_init(argc, argv);
- 
+
+  GRINS::SimulationInitializer initializer;
+
   GRINS::SimulationBuilder sim_builder;
```
roystgnr: Could we have SimulationBuilder create a SimulationInitializer?  Then we wouldn't need to change so much user code.

pbauman: My intent is for SimulationBuilder to go away after properly factorying out Solver, QoI, etc. (goal is to finish that before tomorrow morning). Also, the PR I'm working on right now will hide all this stuff from the user.

pbauman:
Runner PR (#486) hides this away, so I'm going ahead and merging.

PR #485, opened 2017-04-25
Remove GRVY

pbauman:
This removes all traces of GRVY from GRINS. It was never really used heavily, AFAIK, and provides no functionality for timing that we can't get from libMesh directly and would be confusing to anyone not familiar with it. So let's just get it out of the source tree.

I branch from #484, so this shouldn't be merged until that is. Will plan on merging this at the end of today barring any complaints.


pbauman:
Rebased after merging #484. Would like to merge this ASAP.

PR #491, opened 2017-05-01
Format Spacing Script, make reindent

pbauman:
This adds the make target reindent and the accompanying reindent.sh
script. This script is a modified version of the one in libMesh/libmesh.
In particular, the differences are we untabify, remove trailing whitespace, and *do*
indent in the namespace.

Note `make reindent` works from the (out-of-source) build directory.

Now if we could only capture @jwpeterson's regex foo for all the other formatting...


jwpeterson:
> Now if we could only capture @jwpeterson's regex foo for all the other formatting...

Sorry, I don't know how to do text replacement in Emacs' batch mode, plus I never really trusted my regexes enough to script them 😊 

You might be better off using perl for some things because Emacs' regular expressions are... weird. To change `if ( foo )` to `if (foo)` interactively, for example, one can run:
```
tags-query-replace if ( \(.*\) ) RET if (\1)
```
which uses capture groups.

pbauman:
Thanks! I didn't actually mean to imply having to run it through `emacs`, I'm happy to plug some perl lines into that script. My regexp is basically... non-existent, so I was just hoping you might've captured whatever you did in some lines of `perl` (or whatever) to do things like, e.g., `Foo& foo` -> `Foo & foo`.

jwpeterson:
> do things like, e.g., Foo& foo -> Foo & foo.

I used something like the following perl command, which matches any lowercase letter immediately followed by an ampersand and inserts a space, to catch almost all of that particular case.
```
perl -pli -e 's/([a-z])&/\1 &/g' foo.C
```

Fixing asterisk spacing was a little trickier because you will get some false positives in math formulas that use multiplication, unless you also plan to enforce uniform spacing around math operators too.

Issue #495, opened 2017-05-03
[Discussion]: Parsing - Object Constructors or Factories


pbauman:
Wanted to capture some thoughts for discussion while I was thinking about it. Aimed mainly @roystgnr but anyone who's reading can feel free to chime in.

Going through and updating the way we build objects based on `FactoryAbstract` got me thinking more about how we build our objects, in particular with having to parse everything using `GetPot`. Originally (a long time ago, etc.), I had been thinking have every object take `GetPot` and worry about it's own things to avoid the catch-everything-in-input style that we had in FIN-S.

Overtime, the parsing has gotten more complicated and connected since we want to minimize duplication in the input file while at the same time sharing information between modules, e.g. only one `Variables` section in the input file while, at the same time, all the `Physics` classes should be able to understand the variables.

I've been wondering if maybe the parsing work should be done in the `Factory` objects. That is, the objects we're building (`Physics`, `QoI`, `Solver`, etc.) should be "pure" in the sense that they get handed everything they need to start working (perhaps needing an `init()` call before computation, but otherwise no parsing or anything): pass by value for constants, pass by (const) reference for things the object doesn't own (e.g. variables are maintained in the `VariableWarehouse` since lots of things use the variables), shared_ptr for things that have shared ownership and could potentially be NULL (I can't think of anything off-hand that fits this bill), and `unique_ptr` for things it will own.

Obviously, such purity cleans up the guts of our "core" objects and moves the parsing complexity to the factory, which is both good and bad. The good is easier testing and understanding of the core objects (both for users and developers) and we separate all the complicated parsing semantics away from the functionality of the core object. It would also move the parsing out of the parameter registration: the `Physics` constructor, for example, could just directly register the objects that it's handed and not worry about the parsing or error checking issues.

The bad is that each leaf object may have slightly different things it needs so then we end up with nearly 1 factory for every object. That does sound bad, but I wonder if maybe that can also be mitigated to some degree because now things like multiple inheritance makes sense for the factories. For example, instead of `VariableParsing` being static-all-the-things, it becomes non-static and then `PhysicsFactory` and `QoIFactory` now also inherit from `VariableParsing` because they are variable parsers (among other things). So maybe we can gain a lot more contained reuse in the factories to help offset the (likely) bloat in the number of leaf classes for the factories.

[...]

PR #496, opened 2017-05-17
Refactoring Antioch Interface, Part 1

pbauman:
The ultimate goal is to be able to use any of the thermo models in Antioch; right now we're only using CEA. This is part 1.

First part is now templating many of the Antioch interface classes on the `KineticsCurveFitType`, i.e. we're using `Antioch::NASAThermoMixture` etc. kinetics wrapper. This templating then filtered down the hierarchy. Currently, we're still only instantiating `CEACurveFitType`. A future PR will enable NASA9 (basically CEA, just different parsing) and NASA7.

I wanted to do this as a separate PR because I want to redo the construction of the AntiochMixture objects inline with discussions in #495 before updating the parsing of the Antioch thermo models and that plus this would be a pretty large PR.


pbauman:
Rebased on master after #500.

PR #498, opened 2017-05-22
Cleanup Warnings

pbauman:
A little tidying up of warnings. If I turn of Cantera, this produces no warnings for me and, with a little build system work, we'll be able to start turning on `-Werror` on `femputer` builds.

I'll merge this very soon because I want to use this to test new `femputer` recipes I added for merges-to-master.


pbauman:
The only failure is the expected one, for which there is an upstream fix, so going ahead and merging this since no runtime behavior is changed and this will allow me play/fix the merge-to-master `femputer` recipes.

PR #501, opened 2017-05-26
Fix builds with Antioch disabled

roystgnr:
We should probably get an antioch-free build into the GRINS Civet recipes.


pbauman:
Dammit, sorry. Thanks for the quick fix. I'll try and get a recipe for that going today before I leave, after which I'll merge.

pbauman:
OK, have `femputer` recipes added for both PR testing and merges to master. I've abused #472 to make sure we're capturing the failure before this was merged. Going ahead and merging this and will fixup anything that needs to be fixed up in the recipe for the `nooptdeps`.

PR #502, opened 2017-06-06
Refactor Antioch Wrapper Construction

pbauman:
This moves the construction of Antioch mixture objects to builder objects to try and better contain the parsing. The new constructors now take unique pointers to the necessary things that the object needs to run; those are constructed in the builder. Also updated tests in the appropriate places to use the new construction objects.

I've left the deprecated constructors in for now, but we'll remove them after the next release. I have a branch where I've removed them to make sure we've cleaned out the deprecated constructor usage internally (I'll open a separate PR for that).

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -472,6 +472,7 @@ include_HEADERS += properties/include/grins/materials_parsing.h
 include_HEADERS += properties/include/grins/viscosity_base.h
 include_HEADERS += properties/include/grins/antioch_thermo_curve_fit_instantiation_macro.h
 include_HEADERS += properties/include/grins/antioch_options_naming.h
+include_HEADERS += properties/include/grins/antioch_mixture_builder_base.h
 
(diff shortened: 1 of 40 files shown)


pbauman:
Squashed in fix for "nooptdeps" build (thanks femputer!).

roystgnr:
+1,473 -109

You're *sure* this is refactoring to "contain" code?  ;-)

pbauman:
lol :P There's -885 with the deprecated code removed (another branch for which I'll open a PR) and 367 lines of the + is unit testing of the new code, so it's not as bad as it seems. It's those new constructors with arguments on each line because we templated the shit out of Antioch and I like long, descriptive names. :P

pbauman:
Should mention this is an example of trying to implement what I was thinking in #495. You end up with lots of constructor arguments, but I'm liking the user/developer just seeing factory or builder object, and not having all the parsing inside the object itself.

PR #503, opened 2017-06-06
Make CatalycityBase derived from ParameterUser

tradowsk:
Makes `CatalycityBase` a subclass of `ParameterUser` to allow for adding catalycity parameters for sensitivity analysis.

Requires updating `MultiphysicsSystem::register_parameter()` to search the `NeumannBC`s for the params as well, and making `NeumannBCAbstract` a subclass of `ParameterUser` as well

### DIFF
--- a/src/boundary_conditions/include/grins/arrhenius_catalycity.h
+++ b/src/boundary_conditions/include/grins/arrhenius_catalycity.h
@@ -49,6 +49,8 @@ namespace GRINS
       and worry about memory management. */
     virtual CatalycityBase* clone() const;
 
+    virtual void set_parameters(const GetPot & input, const std::string & param_base);
+
   protected:
 
     libMesh::Real _gamma0;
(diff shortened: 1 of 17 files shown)


roystgnr:
Code looks good.  You might want to throw in a regression test ASAP, even if a proper test will be more time (or never) on its way.

pbauman:
Or even just turn on the sensitivity calculation in the existing reacting flow + catalyticity tests.

tradowsk:
I'll add something in a bit

tradowsk:
Added calls for sensitivity analysis to existing RLMNS+catalycity input files. Covers all 3 catalycity types with all their respective params.

I also verified myself that these tests will fail without the changes in e7487ea

pbauman:
OK, that's better than nothing. Still need to add capability to compare against "gold" values, but we don't have that for sensitivities yet, just QoI values.

PR #504, opened 2017-06-07
RayfireMesh::init() fixes for restart

tradowsk:
Issues arose when calling `RafireMesh::init()` on an already refined mesh from a restart file.

The first case was that `INACTIVE` elements were being added to the rayfire. `Elem::neighbor_ptr()` gives the neighbor that is on the same refinement level as the current elem, whether or not that neighbor is `ACTIVE`. Instead, we have to check that the neighbor is active, and if it isn't, we check it's children to try and find the next elem for the rayfire. We also add asserts to try and catch these inactive elements.

The second case is not easily reproduced because it relies on the numbering of elem sides and edges as well as the coordinate values. If a refined elem shares its bottom edge with an unrefined elem and the rayfire traverses along that edge, then the rayfire may wander into the middle of the lower elem and get confused.  This happens with `QUAD9` elements when the intersection point is the middle node of the bottom edge. We add a check which says that if the rayfire `start_` and `intersection_point` are both on an edge, then they must be the end nodes (0 & 1) of that edge elem, and not the middle node if it is an `EDGE3`.  Sorry if this isn't clear, it's much easier to draw out. When I wrote this fix, I was unable to create a unit test that captures this behavior. But I can try again if you think it's necessary.

### DIFF
--- a/src/qoi/include/grins/rayfire_mesh.h
+++ b/src/qoi/include/grins/rayfire_mesh.h
@@ -175,7 +175,7 @@ namespace GRINS
     bool check_valid_point(libMesh::Point& intersection_point, libMesh::Point& start_point, const libMesh::Elem& edge_elem, libMesh::Point& next_point);
 
     //! Knowing the end_point, get the appropraite next elem along the path
-    const libMesh::Elem* get_correct_neighbor(libMesh::Point& end_point, const libMesh::Elem* cur_elem, unsigned int side, bool same_parent);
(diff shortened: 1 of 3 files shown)


tradowsk:
I just rebased this to `grinsfem/master`, is there any way we can merge this soon?

I'm hitting an issue on another branch that should be fixed by these changes, and rebasing that branch onto this was giving me all kinds of unnecessary conflicts.

Also, is `femputer` down?

pbauman:
OK once I can get `femputer` rebooted and run this through it.

pbauman:
All jobs on a09c27e : invalidated by @pbauman

Invalidated to rerun this PR after restarting femputer

PR #505, opened 2017-06-07
AbsorptionCoeff derivatives

tradowsk:
Implementation of analytical derivatives of the Beer-Lambert Law for use in AMR.

We subclass `libMesh::FEMFunctionBase` to provide the `derivatives()` function so `IntegratedFunction` can be used with AMR.

The analytical derivatives are fairly long and nasty. I wrote the derivatives for the lowest-level quantities (i.e. broadening half-widths, Voigt lineshape, linestrength, and partition functions) and then used the product rule to combine these, rather than writing out the full derivatives for everything. The relevant state variables are pressure, temperature, and mass fraction of the species of interest.

### DIFF
--- a/src/qoi/include/grins/fem_function_and_derivative_base.h
+++ b/src/qoi/include/grins/fem_function_and_derivative_base.h
@@ -1,0 +1,54 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// GRINS - General Reacting Incompressible Navier-Stokes
+//
+// Copyright (C) 2014-2016 Paul T. Bauman, Roy H. Stogner
+// Copyright (C) 2010-2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+
+#ifndef GRINS_FEM_FUNCTION_AND_DERIVATIVE_BASE_H
+#define GRINS_FEM_FUNCTION_AND_DERIVATIVE_BASE_H
+
+// libMesh
+#include "libmesh/fem_function_base.h"
+
+// GRINS
+
+namespace GRINS
+{
+  /*!
+    Extends libMesh::FEMFunctionBase by adding a function for computing derivatives
+  */
+  template<typename Output>
+  class FEMFunctionAndDerivativeBase : public libMesh::FEMFunctionBase<Output>
+  {
+  public:
+
+    //! Function Derivative Evaluation
+    virtual void derivatives( libMesh::FEMContext & context,
+                              const libMesh::Point & p,
+                              const libMesh::Real & JxW,
+                              const unsigned int qoi_index,
+                              const libMesh::Real time = 0.) = 0;
+
+  };
+
+}
+#endif //GRINS_FEM_FUNCTION_AND_DERIVATIVE_BASE_H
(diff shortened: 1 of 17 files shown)


src/qoi/include/grins/fem_function_and_derivative_base.h
```diff
@@ -0,0 +1,65 @@
...
+  */
+  template<typename Output>
+  class FEMFunctionAndDerivativeBase : public libMesh::FEMFunctionBase<Output>
+  {
+  public:
+
+    //! Function Evaluation
+    virtual Output operator()( const libMesh::FEMContext & context,
+                                      const libMesh::Point & qp_xyz,
+                                      const libMesh::Real t = 0.) = 0;
```
pbauman: Why is this function here? This is already pure virtual in `FEMFunctionBase`

src/qoi/include/grins/fem_function_and_derivative_base.h
```diff
@@ -0,0 +1,65 @@
...
+                              const libMesh::Point & p,
+                              const libMesh::Real & JxW,
+                              const unsigned int qoi_index,
+                              const libMesh::Real time = 0.) = 0;
+
+    //! Vector of Function Evaluations
+    virtual void operator()( const libMesh::FEMContext & context,
+                             const libMesh::Point & p,
+                             const libMesh::Real time,
+                             libMesh::DenseVector<Output> & output) = 0;
```
pbauman: Same here.

tradowsk:
I don't have any test cases outside of qualitative validation from various runs that the AMR seems to be targeting the right areas.

I don't have any "gold data" against which to compare, and running the same equations through MATLAB or something won't really show anything.

I'm open to ideas though

pbauman:
What about a CppUnit-based test where you finite difference w.r.t. the state variable and compare your computation?

src/qoi/include/grins/fem_function_and_derivative_base.h
tradowsk: Sorry, I had a `stash` hanging around that got rid of that. Will squash it in

tradowsk:
Squashed in getting rid of the unneeded functions in `FEMFunctionAndDerivativeBase`

tradowsk:
> What about a CppUnit-based test where you finite difference w.r.t. the state variable and compare your computation?

I'll work on getting something like that going today

tradowsk:
I've run into an issue with the test I have been working on. To do the finite difference, I was hoping to use the command-line override functionality instead of adding a bunch of input files.

If I try to use `Simulation` directly, it doesn't change anything because the `command_line` parameter is [commented out](https://github.com/grinsfem/grins/blob/master/src/solver/src/simulation.C#L98).

If I try and use the `Runner`, I trip [this libMesh assert](https://github.com/libMesh/libmesh/blob/master/src/base/libmesh.C#L350). The runner code in the last commit (which will fail) was taken from my 701 project and I had no issues with it there. So it must be something with `CPPUnit` then?

Relevant stack trace:
````
#3  0x00007f4ce657a325 in libMesh::print_trace (out_stream=...) at /leela1/data/shared/software/sourcesdir/libmesh/libmesh-master/src/base/print_trace.C:192
#4  0x00007f4ce6574a41 in libMesh::MacroFunctions::report_error (file=0x7f4ce73deee8 "/leela1/data/shared/software/sourcesdir/libmesh/libmesh-master/src/base/libmesh.C", line=350, date=0x7f4ce73de8f1 "May 19 2017", time=0x7f4ce73de8e8 "10:24:17") at /leela1/data/shared/software/sourcesdir/libmesh/libmesh-master/src/base/libmesh_common.C:82
#5  0x00007f4ce655a7fe in libMesh::LibMeshInit::LibMeshInit (this=0x28e7450, argc=2, argv=0x7fffa7099a20, COMM_WORLD_IN=1140850688) at /leela1/data/shared/software/sourcesdir/libmesh/libmesh-master/src/base/libmesh.C:350
#6  0x00007f4ce824fa9f in GRINS::Runner::Runner (this=0x7fffa70997a0, argc=2, argv=0x7fffa7099a20) at ../../src/src/solver/src/runner.C:41
#7  0x00000000004c32a2 in GRINSTesting::SpectroscopicAbsorptionTest::test_T_derivatives (this=0x23009a0) at ../../src/test/unit/spectroscopic_absorption_test.C:112
````

tradowsk:
Also weird that femputer shows a pass on `devel-nooptdeps` when I should still trip the libMesh assert. Even if I didn't, I'm trying to set the Temp BC to -100K, which I would assume throws an error

tradowsk:
Going to split this into 2 smaller PRs

PR #506, opened 2017-06-26
Removing deprecated libMesh calls

tradowsk:
Had these floating around, figured I should do a PR before they get lost.

These commits remove deprecated libMesh function calls in `RayfireMesh` (and associated tests) and replace them with their preferred versions.


pbauman:
Thanks!

PR #507, opened 2017-10-13
Eliminate deprecated code usage

roystgnr:
This gets rid of everything that libmesh/libmesh#1452 catches at compile time, at least.


pbauman:
Awesome, thanks!

PR #508, opened 2017-11-07
HITRAN bugfix

tradowsk:
Main fix is where `HITRAN::T_index()` returned an index that was too high. It should return the index corresponding to the next data value higher than the given `T` but it was actually giving 2 indices higher for most cases.

Should use `std::floor(...) + 1` in all cases. Added additional tests with answer values obtained from `MATLAB interp1()` for verification. I also upped the tolerance for all tests because the previously returned incorrect values differed in the 8th digit and would not fail if using `libMesh::TOLERANCE = 1e-6`

Also updated private function names to match conventions and better describe their functionality.

### DIFF
--- a/src/qoi/include/grins/hitran.h
+++ b/src/qoi/include/grins/hitran.h
@@ -130,10 +130,10 @@ namespace GRINS
     std::vector<std::vector<libMesh::Real>> _qT;
 
     //! Find the index into _T corresponding to the given temperature
-    int _T_index(libMesh::Real T);
+    int T_index(libMesh::Real T);
 
     //! Search through the partition function data to get value for given temperature
-    libMesh::Real search_partition_function(libMesh::Real T, unsigned int iso);
+    libMesh::Real get_partition_function_value(libMesh::Real T, unsigned int iso);
 
     //! Linear interpolation helper function
     libMesh::Real interpolate_values( int index_r, libMesh::Real T_star, const std::vector<libMesh::Real> & y) const;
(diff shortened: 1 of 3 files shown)


tradowsk:
Is there something wrong with femputer? I pass `dbg` on leela...

pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/640/) on bffc6d7 : invalidated by @pbauman

Invalidating to see if segfault in unit_driver was one-off.

pbauman:
@tradowsk I see the failure already. I'm guessing you're using a `dbg` build that doesn't have GLIBCXX_DEBUG going (yes, I need to get a helper function to dump this output into the CIVET log to be viewable from the web):

<pre>
Error: attempt to subscript container with out-of-bounds index 2001, but
container only holds 2001 elements.

Objects involved in the operation:
    sequence "this" @ 0x0x1faec70 {
      type = std::__debug::vector<double, std::allocator<double> >;
    }
FAIL unit_driver (exit status: 134)
</pre>

You ought be to able to trip that in valgrind as well, but I'd suggest having a GLIBCXX_DEBUG libMesh around (should be available in swenv modules, build_libmesh_master_dbg.sh).

tradowsk:
Do I have access to that log file? All I see on Details is the terminal output

pbauman:
What I posted is all that was in the log file. The dbg container is catching out-of-bounds errors. You're going to need to catch that exception in a debugger and work back from there.

tradowsk:
Right, but until you posted that I had no clue why it failed since I couldn't see the log file. Would be good in the future so I don't need to bug you if something like this happens again

pbauman:
Which is why I said "(yes, I need to get a helper function to dump this output into the CIVET log to be viewable from the web)"

tradowsk:
Sorry, must have read over that. Apologies

tradowsk:
I squashed in the fix for that error, and looks like everything is working now

PR #509, opened 2017-11-27
AbsorptionCoeff component derivatives

tradowsk:
This is part 1/2 of closed PR #505

Adds derivatives for each component of `AbsorptionCoeff`. Note this will not support AMR yet, as it does not contain the total QoI derivative. This will be added in a future PR.

For testing purposes, we created a subclass of `AbsorptionCoeff` named `AbsorptionCoeffTesting`. This class `friend`s the `SpectroscopicAbsorptionTest` class to allow access to the `protected` class functions of `AbsorptionCoeff`. We compare a central finite difference approximation of each component to its analytical derivative with respect to temperature, (hydrostatic) pressure, and mass fraction of the species of interest.

### DIFF
--- a/src/utilities/include/grins/physical_constants.h
+++ b/src/utilities/include/grins/physical_constants.h
@@ -33,12 +33,12 @@ namespace GRINS
   namespace Constants
   {
     /*!
-     * Universal Gas Constant, R, expressed in J/(kmol-K)
+     * Universal Gas Constant, R, [J/(kmol-K)]
      */
     const libMesh::Real R_universal = 8314.4621;
 
     /*!
-     * Avogadro's number, particles per mole.
+     * Avogadro's number, [particles/mol]
      */
     const libMesh::Real Avogadro = 6.02214179e23;
 
@@ -51,6 +51,20 @@ namespace GRINS
      * Speed of light in a vacuum, [m/s]
      */
     const libMesh::Real c_vacuum = 299792458;
+
+    /*!
+     * 1 atmosphere in Pascals, [Pa/atm]
+     */
+    const libMesh::Real atmosphere_Pa = 101325.0;
+
+    /*!
+     * Second Radiation Constant hc/k, [m K]
+     * 
+     * h = Planck's Constant
+     * c = speed of light in a vacuum
+     * k = Boltzmann Constant
+     */
+    const libMesh::Real second_rad_const = 1.43877736e-2;
 
 
   } // namespace Constants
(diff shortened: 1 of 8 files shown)


src/qoi/include/grins/absorption_coeff_testing.h
```diff
@@ -0,0 +1,63 @@
...
+
+// GRINS
+#include "grins/absorption_coeff.h"
+
+namespace GRINSTesting
+{
+  class SpectroscopicAbsorptionTest;
+}
+
+namespace GRINS
```
pbauman: I would keep this in the `GRINSTesting` namespace

src/utilities/include/grins/physical_constants.h
```diff
@@ -52,6 +52,16 @@ namespace GRINS
      */
     const libMesh::Real c_vacuum = 299792458;
 
+    /*!
+     * 1 atmosphere in Pascals, [Pa/atm]
+     */
+    const libMesh::Real atmosphere_Pa = 101325.0;
+
+    /*!
+     * Second Radiation Constant, [m K]
```
pbauman: Can you elaborate a bit more here. I'm not sure what constant this is.

src/qoi/include/grins/absorption_coeff_testing.h
tradowsk: So move it to `test/unit/` then as well?

pbauman:
Other than those two nitpicks, this is good with me. Once those are squashed in and tests pass, I'll merge.

src/qoi/include/grins/absorption_coeff_testing.h
pbauman: Yes, please.

pbauman: In fact, you ought not need a separate file, and just park that in place of `#include "absorption_coeff_testing.h"`

tradowsk: Or keep it in `src/qoi/`?

tradowsk:
I squashed in fixes. I put `AbsorptionCoeffTesting` in `test/interface/`.

Pre-squash, had a compilation error in `devel-nooptdeps` that said `AntiochChemistry` is not a member of `GRINS`. Doesn't make sense to me, but we shall see if it's still there when these tests finish

pbauman:
Ah, yeah, I can see it now. You'll need to `#ifdef` around the test that uses `AntiochChemistry` (or better yet, template the test on the chemistry type and then `#ifdef` around the test instantiation). the nooptdeps is testing when there are no optional dependencies, so that build won't have Antioch.

tradowsk:
I went with the easier option.

I'll go back later and be more elegant once the derivative stuff is finally sorted out

pbauman:
Great, thanks!

PR #511, opened 2017-12-04
Ignore generated files.

jwpeterson:
After bootstrapping and configuring GRINS, I end up with a ton of untracked files. This commit ignores them all, but I'm surprised no one has been annoyed by this before and I therefore might be missing something...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,21 @@ examples/Makefile.in
 src/Makefile.in
 test/Makefile.in
 
+.deps
+*.lo
+*.o
+.dirstamp
+Makefile
+config.log
+config.status
+_configs.sed
+examples/*/run.sh
+.license.stamp
+grins_config.h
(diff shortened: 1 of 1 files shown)


roystgnr:
I guess we've always been doing out-of-source builds, for which only the top build directory shows up in the untracked file list.

roystgnr:
:+1: anyway; we shouldnt' be *forcing* out-of-source builds on people.

jwpeterson:
Oh shoot, I was just following the README.md instructions, which didn't mention anything about `mkdir build` that I saw, so that must be it.

pbauman:
Yes, I've always done out-of-source builds. There's nothing preventing doing in-source builds, I just think it's poor styles. Nevertheless, I'm fine with ignoring the cruft in an in-source build.

pbauman:
Thanks @jwpeterson

PR #514, opened 2017-12-08
Include libmesh header that provides the UniquePtr alias.

jwpeterson:
In a separate commit, I will simply update all the UniquePtr instances
in GRINS to std::unique_ptrs.

Refs libMesh/libmesh#1525.


roystgnr:
This is pretty awesome, but I stand by my attitude in https://github.com/libMesh/libmesh/pull/1525 - requiring user codes to make these changes is still API breakage, and I'd be astonished if GRINS was the only code affected by it.

roystgnr:
:+1: here, regardless

PR #516, opened 2017-12-08
Use std::unique_ptr instead of UniquePtr alias.

jwpeterson:
This can and should be merged in addition to #514 even though it
technically obviates the former.


roystgnr:
Looks good except for the indentation breakage.

jwpeterson:
Do you guys have a script or editor settings or something? Me emacs seems to want to indent everything differently...

pbauman:
> Do you guys have a script or editor settings or something?

`make reinvent` should work. It's the same as libMesh, except I indent within `namespace` blocks` whereas libMesh's doesn't.

roystgnr:
:+1: but I'll wait for @pbauman to see it before we push the button.

pbauman:
Looks great, thanks! I'd like to hold off on merging for 1 day or 2 to try and squeeze out a release so we have a clean breaking point between requiring C++11 and not. Shouldn't take long, main thing will be updating required libMesh versions etc.

pbauman:
Replaced by #522. It seemed easier to just redo the sed command then fix the rebase. Sorry and thanks @jwpeterson.

PR #513, opened 2017-12-04
Fix specialization-after-instantiation errors.

jwpeterson:
This gets GRINS compiling for me with clang, thanks @roystgnr for compiler diagnostic interpretation.

Fixes #510.

PR #517, opened 2017-12-10
Fix remaining clang instantiation warnings.

jwpeterson:
This commit is along the same lines as #513, however these were just warnings while the others were actual compiler errors under clang. Anyone compiling GRINS with clang will probably appreciate having this patch (otherwise there are *a lot* of warnings) but I also think this code is pretty ugly/not maintainable so in the long run some kind of redesign might be the best solution...

These warnings were specific to the static data members of the class
and looked like the following:

../../src/boundary_conditions/include/grins/dirichlet_bc_factory_function_old_style_base.h:88:16: warning:
instantiation of variable 'GRINS::DirichletBCFactoryFunctionOldStyleBase<libMesh::FunctionBase<double> >::_var_names_old_style' required here, but no definition is
available [-Wundefined-var-template]
    if( !this->_var_names_old_style )
               ^
../../src/boundary_conditions/include/grins/prescribed_vector_value_dirichlet_old_style_bc_factory.h:40:5: note:
in instantiation of member function 'GRINS::DirichletBCFactoryFunctionOldStyleBase<libMesh::FunctionBase<double> >::check_state' requested here
    PrescribedVectorValueDirichletOldStyleBCFactory( const std::string& bc_type_name )
    ^
../../src/boundary_conditions/include/grins/dirichlet_bc_factory_function_old_style_base.h:72:44:
note: forward declaration of template entity is here
    static const std::vector<std::string>* _var_names_old_style;
                                           ^
../../src/boundary_conditions/include/grins/dirichlet_bc_factory_function_old_style_base.h:88:16:
note: add an explicit instantiation declaration to suppress this warning if 'GRINS::DirichletBCFactoryFunctionOldStyleBase<libMesh::FunctionBase<double> >::_var_names_old_style'
      is explicitly instantiated in another translation unit
    if( !this->_var_names_old_style )
               ^

### DIFF
--- a/src/boundary_conditions/include/grins/dirichlet_bc_factory_abstract.h
+++ b/src/boundary_conditions/include/grins/dirichlet_bc_factory_abstract.h
@@ -33,6 +33,13 @@
 
 namespace GRINS
 {
+  // Specializations must be declared before any automatic instantiation.
+  template<> MultiphysicsSystem * BCFactoryAbstract<libMesh::DirichletBoundary>::_system;
+  template<> const GetPot* FactoryWithGetPot<libMesh::DirichletBoundary>::_input;
+  template<> const std::set<BoundaryID> * BCFactoryAbstract<libMesh::DirichletBoundary>::_bc_ids;
+  template<> const FEVariablesBase * BCFactoryAbstract<libMesh::DirichletBoundary>::_fe_var;
+  template<> std::string BCFactoryAbstract<libMesh::DirichletBoundary>::_section;
+
   class DirichletBCFactoryAbstract : public BCFactoryAbstract<libMesh::DirichletBoundary>
   {
   public:
--- a/src/boundary_conditions/include/grins/dirichlet_bc_factory_function_base.h
+++ b/src/boundary_conditions/include/grins/dirichlet_bc_factory_function_base.h
@@ -31,6 +31,9 @@
 
 namespace GRINS
 {
+  template <>
+  const FEVariablesBase* BCFactoryAbstract<libMesh::DirichletBoundary>::_fe_var;
+
   template<typename FunctionType>
   class DirichletBCFactoryFunctionBase : public DirichletBCFactoryAbstract
   {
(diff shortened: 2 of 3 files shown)


jwpeterson:
Hmm... the fact that this fixes warnings in clang while *breaking the build* in GCC has me very confused.

pbauman:
All [jobs](https://femputer.eng.buffalo.edu:8000/event/341/) on d775d1f : canceled by @jwpeterson

This breaks the build, no reason to keep testing.

pbauman:
> but I also think this code is pretty ugly/not maintainable so in the long run some kind of redesign might be the best solution...

Suggestions always welcome.

pbauman:
I should have some time this afternoon/evening. Will see if I can reproduce clang warnings on my laptop.

jwpeterson:
> but I also think this code is pretty ugly/not maintainable so in the long run some kind of redesign might be the best solution...

> Suggestions always welcome.

Sorry, my comment may have made it sound like I was referring to the original code as ugly, but I only meant the workaround.

pbauman:
I did indeed think you were referring to the original code. :) No offense taken. Quite the contrary, I always welcome feedback on design improvements. Sorry I didn't have time to check this yesterday.

pbauman:
So I was tinkering with clang on my laptop this weekend and I could reproduce all these warnings. It looks to me like these are really unnecessary because you'll end up with a linker error later if you don't eventually instantiate the objects (which is what this warning is about). I would suggest that we just instead disable that particular warning when using clang. FWIW, when I was googling about this error, I came across [this deal.ii thread](https://github.com/dealii/dealii/issues/3705) where they came to the same conclusion.

hartwork:
@jwpeterson thanks for this pull request: it helped me with making commit https://github.com/Libvisual/libvisual/pull/121/commits/8585057711c81bf18a8fd25ab645b784a8ac875f of pull request https://github.com/Libvisual/libvisual/pull/121 :pray:

PR #518, opened 2017-12-12
SpectroscopicAbsorption QoI element_qoi_derivative

tradowsk:
This PR is part 2/2 of PR #505 

Here we plug into the AMR infrastructure. `FEMFunctionAndDerivativeBase` extends `FEMFunctionBase` to add a new function that supports calculating derivatives with respect to state variables.

`finalize_derivative()` was added to `libMesh` in libMesh/linMesh#1445 which we need to apply the chain rule to our QoI.

We include a CPPUnit test that checks a finite difference approximation of `element_qoi()` against the output of `element_qoi_derivative()`.

There will be 1 more PR that will add a test for `assemble_qoi()` and add support for calculating derivatives with respect to all species in the flow, not just the species of interest as we have here.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -497,6 +497,7 @@ include_HEADERS += qoi/include/grins/qoi_options.h
 include_HEADERS += qoi/include/grins/hitran.h
 include_HEADERS += qoi/include/grins/absorption_coeff.h
 include_HEADERS += qoi/include/grins/spectroscopic_absorption.h
+include_HEADERS += qoi/include/grins/fem_function_and_derivative_base.h
 
 # src/solver headers
 include_HEADERS += solver/include/grins/solver.h
--- a/src/qoi/include/grins/absorption_coeff.h
+++ b/src/qoi/include/grins/absorption_coeff.h
@@ -36,6 +36,7 @@
 #include "grins/single_variable.h"
 #include "grins/multicomponent_variable.h"
 #include "grins/variable_warehouse.h"
+#include "grins/fem_function_and_derivative_base.h"
 
 namespace GRINS
 {
@@ -59,7 +60,7 @@ namespace GRINS
     A chemistry library (Antioch or Cantera) is also required.
   */
   template<typename Chemistry>
-  class AbsorptionCoeff : public libMesh::FEMFunctionBase<libMesh::Real>
+  class AbsorptionCoeff : public FEMFunctionAndDerivativeBase<libMesh::Real>
   {
   public:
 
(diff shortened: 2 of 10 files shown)


pbauman:
@tradowsk Looks like missing headers. Think you can get those fixes in this morning? I'm prepping 0.8.0 release and we could squeeze this in.

pbauman:
@roystgnr Any interest in putting this new `derivatives()` method in 
4f7c1eb back in `libMesh::FEMFunctionBase` at some point? Certainly not immediately, but wanted to put that out there.

@tradowsk added it here so he could keep moving forward and I know that at some point we wanted to include the AD work of INL folks, but could still be useful to have analytical hooks for heinously complicated QoIs like this `SpectroscopicAbsorption`.

tradowsk:
I forgot to add this in the description, but this PR requires libMesh daf2f70d7 for the `finalize_derivative()` function, which includes the necessary header. I was actually going to move commit baaaa1cd to the next PR since the test  I have here doesn't use it.

tradowsk:
As for the 0.8.0 release, if you can wait until this afternoon, I can also put up the final PR in the SpectroscopicAbsorption QoI trilogy, so that it would have full, unit-tested AMR capabilities. I have a sign flip that I need to track down first.

But if you need to push out the release this morning, I would say don't include this PR since technically it gives incorrect results.

tradowsk:
I moved the `finalize_derivative()` commit to another branch, so this one should be good to go now without changing the minimum required libMesh version.

But just for future reference, the results returned from this code will **not** be correct and should not be used for AMR.

roystgnr:
@pbauman - it couldn't be virtual, obviously.  Not sure whether the default should be libmesh_not_implemented() or FDM, though.

pbauman:
> I moved the finalize_derivative() commit to another branch, so this one should be good to go now without changing the minimum required libMesh version.

@tradowsk: Perfect, thanks.

> But just for future reference, the results returned from this code will not be correct and should not be used for AMR.

OK.

> @pbauman - it couldn't be virtual, obviously. Not sure whether the default should be libmesh_not_implemented() or FDM, though.

@roystgnr: You mean pure virtual, correct? I think it needs to at least be virtual. I think FDM would be a good default to provide. If you're interested we can think about porting this to libMesh, but it may not be high priority, particularly if there's only luke warm interest.

pbauman:
OK, merging. This will be the last PR for 0.8.0. I'll update CHANGES and push a release.

roystgnr:
Pure virtual; obviously it needs to be virtual.  Sorry about the brain fart.

FDM might be reasonable if we scaled the perturbations by the size of the current element; it's not like doing parameter sensitivity FDM where we basically depend on the users to manually tweak stencil size.

PR #520, opened 2017-12-15
SpectroscopicAbsorption assemble QoI Derivatives

tradowsk:
Finally, at long last, we have commit 3/2 of PR #505

Here we implement the `finalize_derivative()` function in `QoIBase` and `CompositeQoI`. By default, these implementations do nothing. This commit requires libMesh daf2f70

We then add a non-empty definition of `finalize_derivative()` to SpectroscopicAbsorption to give the correct derivative of the exponential QoI.

AbsorptionCoeff is updated so that derivatives are calculated with respect to all species within the flow, not just the absorbing species of interest.

Lastly, we add a CPPUnit test to verify the total QoI `assemble_qoi_derivative()` computation by comparing with a finite difference approximation.

With this PR, the SpectroscopicAbsorption QoI should give correct results for AMR.

### DIFF
--- a/src/qoi/src/spectroscopic_absorption.C
+++ b/src/qoi/src/spectroscopic_absorption.C
@@ -63,4 +63,18 @@ namespace GRINS
     QoIBase::_qoi_value = sys_qoi;
   }
 
+  void SpectroscopicAbsorption::finalize_derivative(libMesh::NumericVector<libMesh::Number> & derivatives, std::size_t qoi_index)
+  {
+    if (!derivatives.closed())
+      derivatives.close();
+
+    // We recalculate the _qoi_value to make sure
+    // it is set and up-to-date
+    libMesh::QoISet qs;
+    qs.add_index(qoi_index);
+    _multiphysics_system->assemble_qoi(qs);
+
+    derivatives.scale(-100.0 * QoIBase::_qoi_value);
+  }
+
 } //namespace GRINS
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,6 @@
 Version 0.9.0 (in progress)
    * C++11 compiler now required
+   * Minimum libMesh git hash for this release is: daf2f70
 
 Version 0.8.0
    * Minimum libMesh git hash for this release is: 8be0a4e
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ In addition to a C++11 compiler, GRINS requires an up-to-date installation of th
 
 libMesh
 -------
-GRINS development both drives and is driven by libMesh development. Thus, the required minimum master hash of libMesh may change in GRINS master. The current required libMesh master hash is 8be0a4e, as of GRINS [PR #499](https://github.com/grinsfem/grins/pull/499).
+GRINS development both drives and is driven by libMesh development. Thus, the required minimum master hash of libMesh may change in GRINS master. The current required libMesh master hash is daf2f70, as of GRINS [PR #520](https://github.com/grinsfem/grins/pull/520).
 GRINS release 0.5.0 can use libMesh versions as old as 0.9.4. Subsequent to
 the 0.5.0 release requires at least libMesh 1.0.0.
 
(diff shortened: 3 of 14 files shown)


tradowsk:
If you want to include this in the 0.8.0 release, feel free. The CIVET testes will all fail to build without that minimum libMesh version update though.

pbauman:
Nah, this will go in post-0.8.0. I'll update CIVET after the release and invalidate the runs.

tradowsk:
> Nah, this will go in post-0.8.0. I'll update CIVET after the release and invalidate the runs.

Sounds good

src/qoi/src/spectroscopic_absorption.C
```diff
@@ -63,4 +63,12 @@ namespace GRINS
     QoIBase::_qoi_value = sys_qoi;
   }
 
+  void SpectroscopicAbsorption::finalize_derivative(libMesh::NumericVector<libMesh::Number> & derivatives)
+  {
+    if (!derivatives.closed())
+      derivatives.close();
+
+    derivatives.scale(-100.0 * QoIBase::_qoi_value);
```
tradowsk: Found a problem here: apparently the adjoint residual error estimator doesn't compute the QoI value before doing the error estimate, meaning that `QoIBase::_qoi_value == 0` and we converge without doing any AMR.

I can reorder the call sequence in `MeshAdaptiveSolver` to calculate the QoI value before doing the error estimate, but it feels a bit hacky to me.

I'm unsure of the proper way to modify the AMR code to account for when the QoI error estimate depends explicitly on the QoI value.

tradowsk: Found a problem here: apparently the adjoint residual error estimator doesn't compute the QoI value before doing the error estimate, meaning that `QoIBase::_qoi_value == 0` and we converge without doing any AMR.

I can reorder the call sequence in `MeshAdaptiveSolver` to calculate the QoI value before doing the error estimate, but it feels a bit hacky to me.

I'm unsure of the proper way to modify the AMR code to account for when the QoI error estimate depends explicitly on the QoI value.

pbauman: Capturing discussion from meat space. I told @tradowsk he should cache a reference to the `MultiphysicsSystem` and then called `assemble_qoi` here to populate the QoI value. We'll see later how big of penalty this is and, if need be, we can add hooks to prevent computing the QoI value more than once, but I don't think we'll need to.

tradowsk: Looks like the easiest way to do this is through `IntegratedFunction::init()` since we get a reference to the `MultiphysicsSystem` as a parameter. I can't do it through the `SpectroscopicAbsorption` constructor since `QoIFactory` does not have access to the system. So `IntegratedFunction` would then be the class to cache the `MultiphysicsSystem` reference

pbauman:
@tradowsk I've updated the libmesh-min-grins version on `femputer` and invalidated your builds to test to make that was done correctly. 

Lucky you it looks like I didn't stomp on you with #523. When you rebase and push your changes to the branch, please also add commits updating the CHANGES file and the README file with the new min libMesh hash. Thanks.

tradowsk:
I squashed in the README/CHANGES updates.

Also added in the code to cache the `MultiphyscisSystem` point in the `IntegratedFunction` class so we can call `assemble_qoi()` from within the `SpectroscopicAbsorption::finalize_derivative()` class (based on my previous review in this PR).

Once `femputer` is satisfied this should be good to go and conclude the SpectroscopicAbsorption saga for the timing being.

pbauman:
Great, thanks.

> I squashed in the README/CHANGES updates.

In the future, please just push this as a separate commit.

pbauman:
Oh, oops, my mistake, I see that's what you did. Thanks!

pbauman:
Great. Time to get some results @tradowsk! :)

tradowsk:
Results? I thought I could get my PhD by just writing code ;-p

pbauman:
> Results? I thought I could get my PhD by just writing code ;-p

@roystgnr @jwpeterson @benkirk

PR #522, opened 2017-12-18
libMesh::UniquePtr --> std::unique_ptr

pbauman:
Also ran `make reindent`. This is just redoing what @jwpeterson did in #516 since some conflicts arose and it seemed easier to just redo the sed from master then fixup the rebase.

This will be the first PR after 0.8.0 release to require C++11.


pbauman:
Going ahead and merging since only changes since successful testing were CHANGES and README

PR #524, opened 2017-12-22
Remove Deprecated Material Parsing

pbauman:
Remove all the code that handled backward compatibility of the original atrocious material property parsing. It's been awhile and there's at least 1 release since deprecation so bye bye. Good riddance.

> 229 additions and 1,105 deletions

*heavy breathing*


pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/921/) on a26a2a8 : invalidated by @pbauman

OK, looks like parsed_qoi.sh heisen-failure again. It&#39;s already passed on #525, so i don&#39;t know what the deal is. Will invalidate to see if it goes away. Cannot repro locally.

pbauman:
`make check` passes in every commit and this is just cleanup so I'm merging right away.

PR #525, opened 2017-12-22
Remove Deprecated Simulation Constructor

pbauman:
It's been deprecated for at least one release. Will rebase once #524 is merged.


pbauman:
OK, I think I have `femputer` fixed. All the tests passed (dbg, devel-nooptdeps, opt all passed, but after fixing CIVET I invalidated devel and it reported back).

PR #526, opened 2017-12-22
Add and Use NonlinearSolverOptions Container

pbauman:
OK, I think I'm starting to figure the right way to contain parsing options. Because I'm an idiot, I've been thinking about `GetPot` as a parser (which it is) rather than a poor-man's-database (which is how we're using it). So, `NonlinearSolverOptions` is basically a very lightweight wrapper around `GetPot` so that we hide the key for which we parse. Then we can dump this object in where ever we need it because it's super lightweight, but we contain the things that might change to one class.

`ErrorEstimatorOptions`, `MeshAdaptivityOptions`, etc. should be refactored in this fashion.

Subsequent PR will change the sectioning for these options. More subsequent PRs will make similar changes to add a `LinearSolverOptions`,  `DisplayOptions`, etc.

Will rebase once #525 is merged (which will be merged once CIVET update snafu is sorted).


pbauman:
And before you comment on the net line gain @roystgnr, the net gain is from *unit testing* this class. So :P

PR #527, opened 2018-01-02
DiffSolver Factory, Support libMesh::PetscDiffSolver

pbauman:
This PR adds the factory pattern for `libMesh::DiffSolver`. Type is now controlled through input, with default going to `libMesh::NewtonSolver`, but we now support and test `libMesh::PetscDiffSolver`.


roystgnr:
Nice!  :+1+

PR #530, opened 2018-01-16
Output after every AMR step

tradowsk:
First attempt at a refactoring of `SteadyMeshAdaptiveSolver` so that the visualization output does not get overwritten on every AMR step.

The new input flag `vis-options/output_every_amr` is set to false by default. Setting it to `true` will output files in the format:  `prefix.amrX.format` (e.g. `test.amr0.xdr`,`test.amr1_mesh.xda` etc.). If you would prefer a different convention, I would be happy to change it.

I also moved the 'meat' of the AMR loop in `SteadyMeshAdaptiveSolver::solve()` to a helper function so that it can be called before and after the AMR has been performed. For example, for `max_refinement_steps = '1'`, you would get an output (QoI value, error estimates, vis, etc.) before the AMR is done and after.  I find this more intuitive than the previous design, where `max_refinement_steps = '1'` wouldn't actually refine/coarsen the mesh.

I haven't done anything in `UnsteadyMeshAdaptiveSolver` yet, so I just put a warning message in there to display if the user sets `output_every_amr = 'true'`.

Like I said, this is a first attempt that meets my specific needs, but I'm open to suggestions.


tradowsk:
Looks like I need to update some of the tests, but I'll wait to do that until we finalize the output design

src/qoi/src/qoi_output.C
```diff
@@ -71,7 +71,7 @@ namespace GRINS
         if( comm.rank() == 0 )
           {
             std::ofstream output;
-            output.open( _file_prefix+".dat" );
+            output.open( _file_prefix+".dat",std::ofstream::app );
```
pbauman: Why force append?

src/solver/src/steady_mesh_adaptive_solver.C
```diff
@@ -93,71 +95,93 @@ namespace GRINS
...
+        context.postprocessing->update_quantities( *(context.equation_system) );
 
-        // Get the global error estimate if you can and are asked to
-        if( _error_estimator_options.compute_qoi_error_estimate() )
-          for(unsigned int i = 0; i != context.system->qoi.size(); i++)
-            {
-              libMesh::AdjointRefinementEstimator* adjoint_ref_error_estimator = libMesh::cast_ptr<libMesh::AdjointRefinementEstimator*>( context.error_estimator.get() );
-              std::cout<<"The error estimate for QoI("<<i<<") is: "<<adjoint_ref_error_estimator->get_global_QoI_error_estimate(i)<<std::endl;
-            }
+        if (context.output_every_amr)
```
pbauman: Instead of a bool, maybe an integer to output every n refinement steps?

src/visualization/src/visualization.C
```diff
@@ -119,6 +119,22 @@ namespace GRINS
...
+    std::stringstream suffix;
+
+    suffix << "amr" << amr_step;
+
+    std::string filename = this->_vis_output_file_prefix;
+    filename+="."+suffix.str();
+
+    this->dump_visualization( equation_system, filename, 0.0 );
+
+    return;
```
pbauman: This was bad style that I'd introduced long ago, but haven't removed from everywhere, but let's not introduce more. For void functions, we just let the closing brace be the end and don't explicitly call `return`.

src/solver/src/steady_mesh_adaptive_solver.C
```diff
@@ -93,71 +95,93 @@ namespace GRINS
...
-          this->steady_adjoint_solve(context);
+    // this will print output the error estimates and/or meshes after the final AMR run
+    if (_mesh_adaptivity_options.max_refinement_steps() > 0)
+      if (! converged)
+        {
+          std::cout << std::endl
+                    << "==========================================================" << std::endl
+                    << "Post-AMR Output " << std::endl
+                    << "==========================================================" << std::endl;
+          this->amr_helper(context,_mesh_adaptivity_options.max_refinement_steps(),false);
```
pbauman: Just checking that I'm reading this correclty: this will ensure we do one final solve after the last refinement step? This was actually missing before, we did an AMR step, but didn't solve to update those new dofs, they were just constrained from the previous solution (which is not entirely egregious, but not as accurate).

pbauman:
I think I'd prefer `test.amr.0.xdr` etc. for the naming scheme (just the extra `.` before the integer).

src/qoi/src/qoi_output.C
tradowsk: I guess I can add an addition `bool append = false` parameter so it can be toggled.

src/solver/src/steady_mesh_adaptive_solver.C
tradowsk: Right: previously a 1-step AMR run wouldn't actually refine the mesh at all, just calculate the errors and output the original mesh if requested. Intuitively, I would assume that a 10-step AMR run would refine the mesh 10 times and give me 11 error estimates and QoI values (from the original mesh and after the 10 refinements)

tradowsk: I do need to make a small modification here so that I can get my QoI error estimates without doing a refinement

src/solver/src/steady_mesh_adaptive_solver.C
tradowsk: I guess that would make sense, especially in the context of the `UnsteadyMeshAdaptiveSolver`

src/visualization/src/visualization.C
tradowsk: Sorry, too much copy-pasta...

tradowsk:
> I think I'd prefer test.amr.0.xdr etc. for the naming scheme (just the extra . before the integer).

I wonder how this scheme will work if we have to append the timestep index as well in the unsteady case. `test.amr.0.0.xdr`?

src/qoi/src/qoi_output.C
pbauman: Hmm. I guess it's not unreasonable to just append QoI values after each AMR step. And if you're not doing AMR, it won't matter. OK, I'm fine with just having it append.

src/solver/src/steady_mesh_adaptive_solver.C
pbauman: Yeah, I was actually thinking about unsteady case, but yeah, for steady case, let's be able to reuse that code and if we're just and if the user put something > 1, we can flash a warning and switch it back to 1.

pbauman:
> I wonder how this scheme will work if we have to append the timestep index as well in the unsteady case. test.amr.0.0.xdr?

Lol. What about when we get around to adding `CheckpointIO`?! `test.amr.0.0.0.dat`Yeah, that'll be a horror show. We'll want to update the unsteady naming to be something like `test.amr.0.ts.0`.

src/qoi/src/qoi_output.C
tradowsk: The only potential problem I can see is that if the file already exists from a previous run and we append to it, it might get confusing.

pbauman: Yeah, I thought about that, but if we're not appending, we're overwriting. Six in one, half dozen the other IMO, so instead of adding a bunch of extra code to worry about the case, let's just carefully document in the `master_example_inputfile.in`.

tradowsk:
Alright, I think I have a decent design here with just some minor tweaks.

I'm going to hold off on those for a while so I can get some runs in and nail down a good test case.

src/solver/src/steady_mesh_adaptive_solver.C
tradowsk: Sorry, not quite sure what you're saying here...

src/qoi/src/qoi_output.C
tradowsk: Sounds good

src/solver/src/steady_mesh_adaptive_solver.C
pbauman: What my stream-of-consciousness comments don't make sense?! :P Let's use integer for steady case, but if the user puts anything other than 0 or 1, we print a warning and change it back to 0 or 1 (whatever makes sense).

tradowsk: Hadn't had my morning tea yet, that makes more sense haha

PR #534, opened 2018-04-03
Spectroscopy QoI example

tradowsk:
A new example for the Spectroscopy QoI.

SpectroscopicAbsorption QoI only calculates absorption value at a single wavenumber. This example runs the QoI in a loop to produce a plot of absorption vs wavenumber for a given range specified in the input file.

`SpectroscopyExample/wavenumber_range` specifies `'nu_min nu_max nu_step'` and `SpectroscopyExample/output_prefix` specifies the prefix for the output file where the data is written.


pbauman:
How about instead of a proper "app", we just make this an example in the examples directory? This seems awfully specialized to ship as an app, but having as an example illustrates building a user app around this capability. Let me know if you need help with the automake foo in examples.

src/qoi/include/grins/absorption_coeff.h
```diff
@@ -78,6 +78,11 @@ namespace GRINS
                     libMesh::Real desired_nu, const std::string & species,
                     libMesh::Real thermo_pressure);
 
+    void set_desired_wavenumber(libMesh::Real new_nu)
```
pbauman: `set_wavenumber` is clear enough I think.

tradowsk:
Okay, I have converted the app to an example.

Example input file uses a 10x10 mesh and has homogeneous conditions for simplicity. Also includes all the necessary spectroscopy data files to scan the wavenumber range [2717.0,2174.0] cm^-1 with a step of 0.01 cm^-1. Run takes about 5 seconds total on my machine.

pbauman:
Please rebase on master since #536 has been merged. Also, the nooptdeps test is failing to build, looks like you're not guarding the code in the new example if there's no Antioch.

tradowsk:
I always forget about guarding for Antioch, sorry...

Would the best way be to guard the body of the example `main()` function and add a `libmesh_error_msg` when not build with Antioch?

pbauman:
> Would the best way be to guard the body of the example main() function and add a libmesh_error_msg when not build with Antioch?

That's fine with me. Just make sure you indicate that Antioch's what missing and GRINS should be configured with Antioch for that example to work.

pbauman:
This is ready to go?

tradowsk:
Yes, was just waiting for femputer to finish (was really slow for some reason)

roystgnr:
> Yes, was just waiting for femputer to finish (was really slow for some reason)

:sweat_smile:

PR #535, opened 2018-04-03
Use a better preconditioner for Poiseuille test

roystgnr:
This prevents the test from breaking in parallel with np > 1 for me
with https://github.com/libMesh/libmesh/pull/1621


roystgnr:
I tested myself on DistributedMesh with np==1 through np==24.  That's not a *bullet proof* test, since ReplicatedMesh will leave us with different partitionings still, but it's pretty good evidence.

PR #540, opened 2018-06-05
Update includes for new enum structure.

jwpeterson:
I'm preparing to forward declare most of the enums in libmesh (instead of #including their headers). This will require some slight changes in dependent apps, but luckily they can be made ahead of time.

Refs libMesh/libmesh#1713.


pbauman:
Fabulous, thanks @jwpeterson

PR #541, opened 2018-06-15
Fixing issue where Cantera and Antioch were returning different units…

klbudzin:
… for the Molecular Weight of a species.

### DIFF
--- a/src/properties/include/grins/cantera_mixture.h
+++ b/src/properties/include/grins/cantera_mixture.h
@@ -135,7 +135,8 @@ namespace GRINS
   libMesh::Real CanteraMixture::M( unsigned int species ) const
   {
     // Cantera returns molar mass in kg/kmol
-    return _cantera_gas->molecularWeight(species);
+    // Need to return kg/mol to match Antioch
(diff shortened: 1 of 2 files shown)


roystgnr:
A factor of a thousand here, a factor of a thousand there, and pretty soon you're talking about real errors!

How does this not trigger any errors from CI?  Do we just not have Cantera coverage?

@pbauman, could this be at the root of the differing behavior you once saw from Cantera vs Antioch on flame problems?

pbauman:
> A factor of a thousand here, a factor of a thousand there, and pretty soon you're talking about real errors!
>
> How does this not trigger any errors from CI? Do we just not have Cantera coverage?

Correct, I never had any "gold" data I trusted and when I did very pedantic testing with Antioch on kinetic rates, I was still getting bogus results from Cantera. 

> @pbauman, could this be at the root of the differing behavior you once saw from Cantera vs Antioch on flame problems?

Almost certainly. I'm going to merge this now since it's an obvious fix and I'm going to work on the thermochemistry interface this morning (for @klbudzin and for @bboutkov) and will see if I can get some reliable tests in. Thanks @klbudzin, good work.

PR #542, opened 2018-06-27
Overlapping map

abhishekmshra:
Updated unit test for fluid_solid_overlapping_map


pbauman:
To add some commentary here.

@abhishekmshra was doing me a favor (and getting his first git cherry-pick assignment) by cherry picking off commits from my abomination of a branch that implements the `OverlappingFluidSolidMap` class (the name sucks, we can change it later). These commits can stand alone, but the idea is that the data structure encapsulates dealing the caching quadrature point <--> elem id mappings for overlapping elements. This will be used subsequently in the `ImmersedBoundary` Physics (once we figure out the bug(s) remaining). But @abhishekmshra grabbed the work I'd done, cherry-picked it off and then added some more unit testing beyond mine.

pbauman:
@abhishekmshra Let's squash together (using `fixup` in the interactive rebase) 
f424929 and bccccd5. You'll then need to "force push" to the branch afterwards.

pbauman:
@abhishekmshra The failures are in dbg with the memory bounds checking on. YOu'll need to load all the `dbg` modules in the environment and then build GRINS with `METHOD=dbg` against the `dbg` modules. It's important to use the `dbg` modules because they have the `-D_GLIBCXX_DEBUG` options enabled that's tripping the failure that will make it *much* easier to debug.

abhishekmshra:
@pbauman Fixed multiphysics_sys.C and parked the commit before OverlappingFluiSolidMap. GRINS-dbg passes now.

pbauman:
Thanks @abhishekmshra 

Could you please add the following to the commit log for 706c8a5 using git rebase and the `reword` option:

"For cases where there are no Physics requested (for testing), we need to guard against calling a nonexistent Physics pointer."

abhishekmshra:
@pbauman Updated the commit log

pbauman:
Fabulous, thanks @abhishekmshra!

PR #543, opened 2018-07-05
Adding a method to calculate species specific, specific heats.

klbudzin:


pbauman:
Thanks @klbudzin. I'm going to merge, but a small nitpick is that we don't use `return;` in `void` functions (I did this a long time ago and haven't cleaned all of them up). It's considered bad style.

PR #544, opened 2018-07-05
second order 'dot var' update

abhishekmshra:


pbauman:
Again, just to add some commentary, this is another PR where I asked @abhishekmshra to cherry-pick off of my working branch implementing IBM. This should be all the changes related to supporting either first-order-in-time or second-order-in-time solvers for second order problems. This relates to the FEMSystem API update way back in libMesh/libmesh#1259.

I'll do one final pass through to make sure we got all the `Physics` classes, then I'll merge.

pbauman:
OK, the RayleighDamping physics need to be updated. I will prepare an amendment commit where we error from those physics if we're using a first order solver and then merge this.

pbauman:
I should add the RayleighDamping is not as trivial as the others because there will be two Jacobian contributions.

PR #546, opened 2018-07-19
Fix ozone.xml Input

pbauman:
@klbudzin pointed out to me a few weeks ago the efficiencies were specified wrong in the XML input file. This PR fixes that and then cleans up that ozone example quite a bit and has a much more detailed description and run scripts. Now runs with both Antioch (constant transport, but AMR at the end!) and Cantera (no AMR yet). Also went ahead and added two new Cantera regression test analogous to existing Antioch versions so we have some Cantera coverage. Pulled this work off of another branch where I'm updating the input parsing for reacting flows.


pbauman:
Squashed those fixes in. Rebasing on master and then will merge after femputer has it's way.

PR #549, opened 2018-07-25
Thermochemsity Parsing Refactoring Part 1: Cantera

pbauman:
Part 1 of a monster branch on refactoring the input parsing for thermochemistry. First is Cantera because was easier to do. 

In short, we've added a new option `chemical_data` that will correspond to the file that contains thermo, transport, and kinetics all in one XML file (this is standard for Cantera). Previously the option lived outside the Cantera section was specifically labeled as kinetics. Secondly, we don't require explicitly (and redundantly) setting the species in the GRINS input file, we now parse the species list from the XML file, based on the gas_mixture specified by the user (since there can be more than 1 mixture in the XML file, which we test). So the new format for Cantera will look like, for example,

<pre>
[Materials]
   [./OzoneGas]
      [./GasMixture]
         thermochemistry_library = 'cantera'

         [./Cantera]
            gas_mixture = 'ozone'
            chemical_data = 'ozone.xml'
         [../]
      [../]
 []
</pre>

I think this is much cleaner. This PR preserves backward compatibility (and spews deprecated warnings) as tests passed before we updated to the new format.

Part 2 will be for Antioch where we will try and get as close to this as we can.

### DIFF
--- a/examples/master_example_inputfile.in
+++ b/examples/master_example_inputfile.in
@@ -176,20 +176,28 @@
         # Species in this gas mixture
         species = 'N2 N C CN'
 
-        # Path to file containing kinetics data.
-        # For Cantera, this may also contain all other thermochemistry data.
-        kinetics_data = 'datafile.xml'
-
         [./Cantera]
-           # Specify the mixture to be used. This is the "mixture"
-           # given in the kinetics_data file. The user must set this
+           # Specify the mixture to be used. This is the "mixture" or "phase"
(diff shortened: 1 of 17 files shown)


pbauman:
I tested with Antioch, but not without Cantera, lol. Was missing one test guard that I just squashed into 
74cbf42. Should be good after that.

pbauman:
OK, I'm going to go ahead and merge now that tests all pass.

PR #550, opened 2018-07-26
Thermochemsity Parsing Refactoring Part 2: Antioch 

pbauman:
Monster PR, but part 2 cousin of #549. This updates the Antioch side. This PR is basically a massive overhaul of the internals of parsing data for Antioch. The code changes have left placeholders for inserting the Anitoch ChemKin parser, but I haven't dropped those in yet. This does add full support for Antioch's XML parser, including not just kinetics, but thermo and transport as well. As such, this PR will push the minimum Antioch version up to  libantioch/antioch@e17822d, as reflected in the README and CHANGES file. We also do not need to explicitly set the species names in the GRINS input file when using XML anymore, we parse it just as we do with Cantera. We do still require it for the ASCII parsing. Finally, the type of curve fit will be deduced based on the input file.

Backward compatibility has been maintained (and tested, didn't update test input files until the end), but didn't add new features to old style so you need to switch (but it's easy).

We determine the Antioch parsing type based on the suffix of the file specified in the new `chemical_data` option: `.dat` is ASCII, `.xml` is XML, and `.chemkin` is CHEMKIN (which isn't implemented yet). For ASCII, you'll still have to set the species names and all the other data.

The new preferred input file format for Antioch specification is identical to Cantera when using XML except
1. We still need the species data file (or default to Antioch's installed file) because Antioch doesn't automatically compute the molecular weights (libantioch/antioch#248)
2. We need set the `thermo_model` to `ideal_gas` (this is the same model as Cantera) or `stat_mech`.
3. We need to set the `transport_model` and associated entries.

So, the new format for using XML input (that includes kinetics, thermo, transport in one file) is

<pre>
[Materials]
  [./2SpeciesNGas]
     [./GasMixture]
        thermochemistry_library = 'antioch'

        [./Antioch]
           chemical_data = './input_files/air_2sp.xml'
           gas_mixture = 'air2sp'
           thermo_model = 'ideal_gas'
           transport_model = 'mixture_averaged'
           viscosity_model = 'blottner'
           thermal_conductivity_model = 'eucken'
           mass_diffusivity_model = 'constant_lewis'
         [../]
   [../]
   [./LewisNumber]
      value = '1.4'
[]
</pre>

For Blottner or Sutherland, you still need the data in either raw txt file or just let it default to Antioch if the species are there (which is what will happen above). For `kinetics_theory` that data is parsed from the XML file.

One should now take the above input and change `antioch` to `cantera` and run with Cantera seamlessly (where Cantera would then be using the transport model specified in the XML file).

@bboutkov @klbudzin This should make your lives quite a bit easier.

I'm pretty happy with this except the `kinetics_theory` model doesn't work like Cantera does. But that's a separate issue. (I get divergence when starting from time 0 in the ozone_flame exampling after a few dozen time steps, and I can't get convergence when restarting from the same file we restart Cantera from to get steady state ozone flame example, dies after 4 or 5 Newton steps. @klbudzin I'm going to want your help in getting to the bottom of that.)

But with this PR we get quite a bit more functionality out of Antioch and it will make it easier to test, etc.


pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/2021/) on 004f77e : invalidated by @pbauman

Updated Antioch version on femputer

pbauman:
Job [GRINS-dbg](https://femputer.eng.buffalo.edu:8000/job/2022/) on 004f77e : invalidated by @pbauman

Updated antioch-dbg version on femputer

pbauman:
That's why we have CI kids. Pushed tolerance tweaks for parallel runs and updated unit test that became stale when optdeps are disabled.

PR #551, opened 2018-08-01
Modularize Displacement Gradient, Stress, and Elasticity Computation in Solids Physics

pbauman:
We're eventually going to want to call these from an ImmersedBoundary type Physics kernel so let's do it separately now. We'll probably want to move to a CRTP interface for SolidMechanicsAbstract to encode this API, but let's actually try and see if this is the right thing to do before enforcing that.


pbauman:
OK, UB firewall is locking femputer out right now. In theory it should be fixed tomorrow. Regardless, whenever it's fixed, I'll redeliver the payload to get the test suite to run.

PR #554, opened 2018-08-03
Better way to generate absorption plot

tradowsk:
This PR generates the same absorption vs. wavenumber plot as the `Spectroscopy` example (#534) but utilizes the `CompositeQoI` class to dramatically speed up this process.

The `Spectroscopy` example loops over _N_ wavenumbers in a given range and recalculates the QoI every time. This results in looping over the entire mesh _N_ times, which is painfully slow for larger meshes or when _N_ is large.

Now, the user can specify `QoI/SpectroscopicAbsorption/desired_wavenumber = 'nu_min nu_max nu_step'` and a separate QoI will be created for each of the _N_ wavenumbers and added to `CompositeQoI`. This means that we only have to loop over the elements **once** and on each element we will calculate the qoi contributions for all _N_ wavenumbers.

Test case: 100x100 QUAD9 element mesh with conditions given in `examples/spectroscopy/spectroscopy.in`
Example code runtime: 165 seconds
New CompositeQoI runtime: 51 seconds

When a user specifies a range of wavenumbers as above, the QoI results will be output in wavenumber,absorption CSV format. This can be directly output to a file for easy plotting by setting `Display/QoI/default_file_prefix = 'prefix'` or can be output to the screen if desired.

The README in the `Spectroscopy` example was updated to indicate this new way for generating an absorption plot, and a sample input file `opt_spectroscopy.in` was included.

### DIFF
--- a/src/qoi/include/grins/absorption_coeff_base.h
+++ b/src/qoi/include/grins/absorption_coeff_base.h
@@ -1,0 +1,69 @@
+//-----------------------------------------------------------------------bl-
+//--------------------------------------------------------------------------
+//
+// GRINS - General Reacting Incompressible Navier-Stokes
+//
+// Copyright (C) 2014-2017 Paul T. Bauman, Roy H. Stogner
+// Copyright (C) 2010-2013 The PECOS Development Team
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the Version 2.1 GNU Lesser General
+// Public License as published by the Free Software Foundation.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc. 51 Franklin Street, Fifth Floor,
+// Boston, MA  02110-1301  USA
+//
+//-----------------------------------------------------------------------el-
+
+
+#ifndef GRINS_ABSORPTION_COEFF_BASE_H
+#define GRINS_ABSORPTION_COEFF_BASE_H
+
+// libMesh
+#include "libmesh/fem_function_base.h"
+#include "libmesh/auto_ptr.h"
+
+// GRINS
+#include "grins/assembly_context.h"
+#include "grins/hitran.h"
+#include "grins/single_variable.h"
+#include "grins/multicomponent_variable.h"
+#include "grins/variable_warehouse.h"
+#include "grins/fem_function_and_derivative_base.h"
+
+namespace GRINS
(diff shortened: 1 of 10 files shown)


src/qoi/include/grins/absorption_coeff_base.h
```diff
@@ -0,0 +1,66 @@
...
+#include "grins/multicomponent_variable.h"
+#include "grins/variable_warehouse.h"
+#include "grins/fem_function_and_derivative_base.h"
+
+namespace GRINS
+{
+  class AbsorptionCoeffBase : public FEMFunctionAndDerivativeBase<libMesh::Real>
+  {
+  public:
+    AbsorptionCoeffBase(libMesh::Real nu)
```
pbauman: I would suggest adding `AbsorptionCoeffBase() = delete;` here to get rid of the default constructor to make sure you don't accidentally end up with a zero wavenumber absorption.

src/qoi/src/qoi_factory.C
```diff
@@ -99,6 +99,9 @@ namespace GRINS
   {
     QoIBase* qoi = NULL;
 
+    // TODO better way to do this? (skip the final add_qoi at the end of the if-else)
```
tradowsk: @pbauman or anyone else, is there a better way to do this?

I have to add the qois to the composite_qoi in a loop at line 243, but there is also a final add at the end of the function (line 258). So if I don't skip that for this case, I would be duplicating the last qoi.

src/qoi/include/grins/absorption_coeff_base.h
tradowsk: I'll add that

tradowsk:
>The example "just works" with run.sh that gets installed?

I didn't change the example code at all, so it works just as it did before. I just updated the README and added an example input file that uses the CompositeQoI way to generate the plot, but that's just for reference and isn't actually used by the example code.

src/qoi/src/qoi_factory.C
pbauman: Yeah, I saw this. The way to actually fix this is make this a proper factory pattern and encapsulate the QoI builds better. But that's beyond the scope of this PR. I can live with this hack for now.

pbauman:
> I didn't change the example code at all, so it works just as it did before. I just updated the README and added an example input file that uses the CompositeQoI way to generate the plot, but that's just for reference and isn't actually used by the example code.

OK, I see. Any reason to keep the old version around now?

tradowsk:
>OK, I see. Any reason to keep the old version around now?

The code is instructive in how to deal with the QoI classes directly, but it's technically redundant with the changes in this PR

tradowsk:
I squashed in the `AbsorptionCoeffBase` constructor deletion, so all that is left is the comment I left you in `qoi_factory`

src/qoi/src/qoi_factory.C
tradowsk: okay, I'll remove the TODO and then we should be set

PR #559, opened 2018-08-08
SpectrosocpicAbsorption Fixups

tradowsk:
This PR makes some minor updates to the `SpectroscopicAbsorption` QoI infrastructure.

We remove constructors from `SpectroscopicAbsorption` and `IntegratedFunction` that simply passed through the `GetPot` input file to the rayfire.

An input file option was added to allow the rayfire mesh to be output: `QoI/[qoi_name]/Rayfire/output_filename`. Note this filename must include a suffix supported by `libMesh::NameBased_IO` (e.g. 'rayfire.e').

Lastly, as a followup to #554 we allow multiple `IntegratedFunction`-based QoIs to share the same rayfire object. Previously, the rayfire was stored as a `std::unique_ptr` in `IntegratedFunction` and was then deep-copied by `clone()`. So if we wanted an absorption plot with 3000 points, we would have 3000 separate rayfire meshes. By storing the rayfire as a `std::shared_ptr` we now always have only 1 rayfire. Checks were added to `RayfireMesh::init()` and `RayfireMesh::reinit()` to ensure those functions are not called multiple times unnecessarily.


pbauman:
Just noting weird compiler behavior for the ether. The earlier failures were because of this constructor:

<pre>
IntegratedFunction<Function>::IntegratedFunction(unsigned int p_level, std::shared_ptr<Function> & f, std::shared_ptr<RayfireMesh> & rayfire, const std::string & qoi_name)
</pre>

`std::shared_ptr` needs a `const` reference to the constructor and we were getting the error for building `_f` without the const. BUT we DIDN'T get a compiler error when trying to build `_rayfire` with the non-const reference `std::shared_ptr`. WTF? I had @tradowsk go ahead and make it const as well in case other compilers beyond GCC 7.2.0 were more discriminating.

tradowsk:
As a followup to @pbauman above, I tried removing the `const`s and switching the order of the parameters:

`IntegratedFunction(unsigned int p_level, std::shared_ptr<RayfireMesh> & rayfire, std::shared_ptr<Function> & f, const std::string & qoi_name)`

and I still got the same error on the `f` parameter.

````
../../src/test/unit/integrated_function_test.C:134:77: error: cannot bind non-const lvalue reference of type ‘std::shared_ptr<libMesh::FunctionBase<double> >&’ to an rvalue of type ‘std::shared_ptr<libMesh::FunctionBase<double> >’
                 integ_func((unsigned int)2, rayfire,f, "integrated_function");
````

So there seems to be something fundamentally different between the 2 objects to make the `f` require a `const` but not the `rayfire`. The obvious difference is that `f` is templated where `rayfire` is not, but I have no clue why this would change anything.

tradowsk:
Can you hold off on merging this for a bit?

I want to double check the logic in the skip reinit() to make sure I got it right.

pbauman:
If you add some libMesh logging there, you ought to be able to see the counts on the function call to that function

tradowsk:
> If you add some libMesh logging there, you ought to be able to see the counts on the function call to that function

I just used a simple `cout` to show when we actually changed the rayfire in reinit, and using `opt_spectroscopy.in` from the Spectroscopy example as a benchmark, it worked just as expected where rayfire reinit only happened once per AMR, not 10 times as before.

I squashed in the removal of that `cout`.

I also looked at the code again and I believe I got the logic correct on the `reinit()` skipping, so once femputer is happy I think this is set.

pbauman:
Looks like femputer is happy. Thanks!

PR #560, opened 2018-08-24
New rayfire 2D solvers

tradowsk:
We reformulate the calculation of the intersection between the rayfire line and the edge of a 2D element. 

When the 2D element is `FIRST` order, all edges are linear so we can actually solve for the intersection point analytically (see `intersection_2D_first_order()`). When the element is `SECOND` order, the edges may be quadratic so we use a Newton iterative solver to find the intersection point (see `intersection_2D_second_order()`)

Two new rayfire tests are included: one is a non-square QUAD4 element that is translated away from the origin, and the other is a QUAD9 where the top edge is quadratic.

To make clear documentation, the `amsmath` package was included in the `doxygen` configuration to allow for the use of `bmatrix` in the comments.

### DIFF
--- a/doxygen/grins.dox.in
+++ b/doxygen/grins.dox.in
@@ -1023,7 +1023,7 @@ PAPER_TYPE             = $(PAPER_SIZE)
 # The EXTRA_PACKAGES tag can be to specify one or more names of LaTeX 
 # packages that should be included in the LaTeX output.
 
-EXTRA_PACKAGES         = 
+EXTRA_PACKAGES         = amsmath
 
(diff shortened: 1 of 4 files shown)


tradowsk:
> We could probably break up the nonlinear part into smaller functions, but I don't think it's necessary.

I agree. I originally hoped to make this more modular but I couldn't figure out a better way than what I have above. Same with the 3D stuff coming soon, at some point we should do a code cleanup and restructure it under the hood.

tradowsk:
I squashed in the fix to the documentation

PR #561, opened 2018-09-05
RayfireMesh 3D support

tradowsk:
Finally, we add support for using a rayfire on a 3D mesh.

The intersection solve is similar to #560. We have an analytic solution for the intersection of a line and a plane for 3D `FIRST` order elements, and then a Newton iterative solve for 3D `SECOND` order elements.

One improvement over #560 is that for second order solves we create a first order element and use the first order solver to identify the side on which the intersection point is located. We then run the second order Newton solve only on this side of the second order element, which saves us from doing a newton solve on every side and trying to identify when a correct intersection point is found.

The CPPUnit tests were refactored to separate 2D and 3D rayfire tests, with the new 3D test mimicking the most important tests from 2D. Current 3D tests cover `HEX` and `TET` elements.


tradowsk:
Um, somehow my TET tests didn't make it into the commits... Hold off on merging this until I figure out where those wandered off to

tradowsk:
Sorry, I pushed my TET code a while ago but somehow it never showed up on GitHub. I rebased to get new hashes and now it seems to have pushed correctly

pbauman:
Thanks!

PR #563, opened 2018-09-25
Rayfire newton solver fix

tradowsk:
Through testing on a large 27k element 3D mesh, we found that the 3D `SECOND` order rayfire intersection point solver did not adequately converge to pass a `libMesh::Elem::contains_point()` check.

This issue is solved by changing the convergence tolerance of the newton step vector l2 norm to `libMesh::TOLERANCE*side_elem->hmax()` based on the checks within `contains_point()` itself. 

However, should the mesh contain large elements of size O(1), we may get a tolerance that is too large. So, we added a floor to the tolerance of `libMesh::TOLERANCE` and use `std::min()` to choose the smaller of the two. This way, smaller elements can have a tighter tolerance to make sure they sufficiently converge, while large elements might not need such strict convergence.

A CPPUnit test was added based on an element from the aforementioned large 3D mesh, hence the seemingly arbitrary node coordinates. Without this PR, the test would fail to converge and would error out.

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -887,7 +887,8 @@ namespace GRINS
           J_inv.vector_mult(delta,F);
 
           // check for convergence
-          if ( delta.l2_norm() < libMesh::TOLERANCE )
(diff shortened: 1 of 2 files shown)


pbauman:
Thanks!

PR #565, opened 2018-12-05
New SpectroscopicTransmission QoI

tradowsk:
This PR is a first step towards adding support for 2D and 3D laser absorption calculations. As part of that, we require the ability to calculate the _spectral transmission_, `T`.

Currently, it is technically possible to use the `SpectroscopicAbsorption` class to get the transmission by `T = 1-A`, where `A = 1-T` is the SpectroscopicAbsorption QoI value (this neglects scattering and emission along the optical path). However, I'm not a fan conceptually of getting a transmission value from an "absorption" class. Secondly, as we move into using this infrastructure with experimental data, we may need to ability to simulate transmission for comparison or statistical reasons. And lastly, if we wish to do AMR based on transmission instead of absorption, the above "math hack" may mess with the signs on the derivatives.

As such, I felt that creating a new standalone QoI specifically for transmission was the clearest way forward. I reworked the spectroscopy QoI src and spectroscopy QoI testing hierarchies to reduce code duplication.


tradowsk:
One question I have is about the 2 new input files I had to add for the transmission cppuint tests, which added over 300+ lines to this PR. The reason is that these new input files use `QoI/SpectroscopicTransmission` instead of `QoI/SpectroscopicAbsorption`; everything else is identical.

Is there a better way of doing input file changes like this so I don't bloat the `test/` dir? I know I can use the `command_line` param to override, but with these I'm adding/dropping entire sections.

tradowsk:
For reference, I can consolidate the input files using the same methodology as PR #566

pbauman:
Sorry to be a PITA, but could you please rebase on master now that I merged #566?

pbauman:
Otherwise looks good. Thanks!

tradowsk:
>Sorry to be a PITA, but could you please rebase on master now that I merged #566?

No worries, I knew it was coming. Branch has been rebased and pushed

PR #566, opened 2018-12-05
Remove unnecessary trivial input files in CPPUnit tests

tradowsk:
This PR cleans up the mess of trivial input files that were included as part of various `RayfireMesh`, `IntegratedFunction`, and `SpectroscopicAbsorption` tests.

We add functions to the `RayfireTestBase` to generate a string to be passed to `GetPot` (as a `std::stringstream`) instead of having a bunch of nearly identical input files floating around.

Other non-trivial files were cleaned up to avoid unnecessary output.


pbauman:
Thanks!

PR #567, opened 2019-01-03
LaserAbsorption QoI class

tradowsk:
Here we add a new QoI that can compute the total absorbed laser intensity across a two-dimensional "laser beam". We use Gauss quadrature to integrate the initial and final laser intensity profiles, and then take their ratio and subtract from 1 to get the absorption QoI value. To calculate the final intensity profile, we needed to do a transmission calculation at each quadrature point.

As such, we needed a QoI object similar to `CompositeQoI` that could store QoI objects internally but was itself derived from `QoIBase`. Hence, `MultiQoIBase` was created as an amalgamation of `QoIBase` and `CompositeQoI`. The `LaserAbsorption` class was then derived from this new class and stores `SpectroscopicTransmission` objects internally.

We also added new classes to evaluate the laser intensity at a given radius from the beam centerline. Currently, we have `ConstantLaserIntensityProfile` for a constant intensity across the beam width, and `CollimatedGaussianLaserIntensityProfile` for when the profile is gaussian but does not change along the optical path (i.e. collimated). New intensity profiles can be added by subclassing the `LaserIntensityProfileBase` class.

This PR does *not* provide support for AMR based on this QoI, so `element_qoi_derivative()` is left as `libmesh_not_implemented()`. This PR also only supports 2D meshes; 3D support will be added in future PR(s).

Intensity profile calculations were validated with MATLAB (since they are simple calculations).

Validation of the LaserAbsorption results was done by comparing with a separate iimplementation of the same calculation that can be found on [this branch](https://github.com/tradowsk/grins/tree/2d_abs) and was included in a recent AIAA journal paper. The absorption results using the same non-trivial test case were identical to the 16th decimal place.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -185,6 +185,7 @@ libgrins_la_SOURCES += properties/src/chemistry_builder.C
 # src/qoi files
 libgrins_la_SOURCES += qoi/src/average_nusselt_number.C
 libgrins_la_SOURCES += qoi/src/qoi_base.C
+libgrins_la_SOURCES += qoi/src/multi_qoi_base.C
 libgrins_la_SOURCES += qoi/src/qoi_factory.C
 libgrins_la_SOURCES += qoi/src/vorticity.C
 libgrins_la_SOURCES += qoi/src/composite_qoi.C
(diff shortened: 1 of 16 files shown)


pbauman:
@tradowsk This is ready to go?

tradowsk:
Yeah, this is good, and once it's merged I can rebase #568 and that is good too.

The issue I have been tracking down is separate from these PRs

PR #568, opened 2019-01-15
LaserAbsorption QoI 3D support and CPPUnit tests

tradowsk:
We add support for using the `LaserAbsorption` QoI on 3D meshes, as an extension to #567. We also add CPPUnit tests on 2D and 3D meshes that use the same flow field as the `spectroscopic` tests, and hence should yield the same QoI values.

For 3D, we use a `QUAD9` elem to define the shape of the circular laser cross-section. We use vector algebra to define the nodes of the QUAD9 from the given 3 points and then attach a quadrature rule. The rest proceeds exactly as in the 2D scenario.

Note that AMR support is **not** included in this PR.

### DIFF
--- a/src/qoi/include/grins/laser_absorption.h
+++ b/src/qoi/include/grins/laser_absorption.h
@@ -34,26 +34,29 @@
 
 namespace GRINS
 {
+  /*!
+    The LaserAbsorption class is a 2D/3D analogue of SpectroscopicAbsorption that
+    can calculate the total absorbed laser intensity of a two-dimensional
+    laser beam across a given flow field.
+    
+    It uses Gauss quadrature to integrate the initial and final laser intensity profiles.
+    
+    Each quadrature point has a corresponding SpectroscopicTransmission object to calculate the
+    transmitted laser intensity along its respective optical path (i.e. RayfireMesh).
+  */
   class LaserAbsorption : public MultiQoIBase
   {
   public:
-
     /*!
-      The LaserAbsorption class is a 2D analogue of SpectroscopicAbsorption that
-      can calculate the total absorbed laser intensity of a two-dimensional
-      laser beam across a given flow field.
-      
-      It uses Gauss quadrature to integrate the initial and final laser intensity profiles.
-      
-      Each quadrature point has a corresponding SpectroscopicTransmission object to calculate the
-      transmitted laser intensity along its respective optical path (i.e. RayfireMesh).
+      Two-dimensional class constructor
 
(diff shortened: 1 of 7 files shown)


tradowsk:
@pbauman I rebased this on master so assuming femputer is happy this is also good to go.

pbauman:
<pre>
/femputer/pbauman/civet_build_testing_root/grins/src/qoi/src/laser_absorption.C:37:34: fatal error: libmesh/communicator.h: No such file or directory
 #include "libmesh/communicator.h"
</pre>

I wonder if we're needing a newer libMesh or if you can rejigger header includes?

tradowsk:
> I wonder if we're needing a newer libMesh or if you can rejigger header includes?

It seems to compile fine on my machine without that header, so I'll push a new file without that header and we'll see if femputer is happy then. Otherwise we would need to bump the min libMesh commit to [`48bd50d`](/libMesh/libmesh/commit/48bd50d4a0e61b4634bbfde366db6ac22930d481)

pbauman:
@tradowsk Looks like header file fix is OK, but need to run make check in parallel? Looks like one of your unit tests died in parallel?

tradowsk:
I'm doing a clean build on my machine to verify and debug

tradowsk:
Clean build with current libMesh `master` passes fine in `dbg` on my machine...

tradowsk:
I guess it's related to this (line 755 in output), but I have no clue what that means or where it's coming from...

```
/femputer/pbauman/swstack/apps/gcc/6.2.0/include/c++/6.2.0/debug/vector:422:
Error: attempt to subscript container with out-of-bounds index 2, but 
container only holds 2 elements.

Objects involved in the operation:
    sequence "this" @ 0x0x28273d8 {
      type = std::__debug::vector<unsigned int, std::allocator<unsigned int> >;
    }
FAIL unit_driver (exit status: 134)
```

tradowsk:
I probably don't see this error on my machine since I'm not running a full `dbg` stack

tradowsk:
So it seems to be erroring out when it tries to access the `w` velocity component, which I forgot to set because the base input file only has `Ux` and `Uy`. shall fix shortly

tradowsk:
Okay I squashed in a change that now passes in `dbg` on my machine using the `glibcxx_debug` flags. Hopefully femputer agrees and this can finally get merged

pbauman:
Job [GRINS-opt](https://femputer.eng.buffalo.edu:8000/job/2920/) on 811c626 : invalidated by @pbauman

Looks like test that has heisenbug that I haven&#39;t ever been able to pin down. Trying again.

PR #569, opened 2019-01-28
Fix Annoyances

tradowsk:
These issues are extremely minor and inconsequential, but they are annoying and clogged up some log files. Particularly getting rid of the deprecated warnings from the spectroscopic tests was a big help.

### DIFF
--- a/src/physics/src/physics_factory_reacting_flows.C
+++ b/src/physics/src/physics_factory_reacting_flows.C
@@ -202,7 +202,7 @@ namespace GRINS
       {
         {
           std::string msg = "WARNING: Specifying thermo_model = "+AntiochOptions::cea_nasa_model();
-          msg += "is DEPREACATED!\n";
+          msg += " is DEPREACATED!\n";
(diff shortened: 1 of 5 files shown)


src/physics/src/physics_factory_reacting_flows.C
```diff
@@ -202,7 +202,7 @@ namespace GRINS
       {
         {
           std::string msg = "WARNING: Specifying thermo_model = "+AntiochOptions::cea_nasa_model();
-          msg += "is DEPREACATED!\n";
+          msg += " is DEPREACATED!\n";
```
roystgnr: Maybe fix the spelling while you're editing anyway?  :-D

roystgnr:
:+1:

src/physics/src/physics_factory_reacting_flows.C
tradowsk: Wow, nice catch! It's fixed in the latest commit

PR #571, opened 2019-02-19
Rayfire 3D fix

tradowsk:
This PR fixes a failure that was observed in a 3D `HEX27` mesh where the rayfire was unable to calculate an intersection point on certain elements. Inspection of these elements revealed that the `HEX8` version had non-linear faces, and so the `RayfireMesh::intersection_3D_first_order(...)` function would fail since it assumed faces were planes.

To address this issue, we drop the `FIRST` order 3D function and instead use the `SECOND` order newton solver function for all 3D elements. This will account for all nonlinearity (since it operates on the elem reference coordinates `xi` and `eta`).

A new CPPUnit test was added for a 3D `HEX27` that previously caused a failure. This fix was also tested on the larger `HEX27` mesh and no failure was found.

### DIFF
--- a/src/qoi/src/rayfire_mesh.C
+++ b/src/qoi/src/rayfire_mesh.C
@@ -1003,165 +1003,145 @@ namespace GRINS
     libmesh_assert_equal_to(cur_elem->dim(),3);
     libmesh_assert(cur_elem->default_order() == libMesh::Order::SECOND);
 
-    // FIXME TA: there is probably a more efficient way to do determine which face has the initial_point
-    // figure out which side has the initial_point so we can skip it above
-    unsigned int skip_side = libMesh::invalid_uint;
-    for (unsigned int s=0; s<cur_elem->n_sides(); ++s)
-      {
-        std::unique_ptr<const libMesh::Elem> side_elem = cur_elem->build_side_ptr(s);
-
-        // using the default tol can cause a false positive when start_point is near a node,
-        // causing this loop to skip over an otherwise valid side to check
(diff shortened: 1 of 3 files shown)


src/qoi/src/rayfire_mesh.C
```diff
@@ -1349,4 +1237,35 @@ namespace GRINS
...
   }
 
+  bool RayfireMesh::system_solve_3x3(libMesh::DenseMatrix<libMesh::Real> & A, libMesh::DenseVector<libMesh::Real> & b, libMesh::DenseVector<libMesh::Real> & x)
+  {
+    bool has_soln = false;
+    libMesh::Real det =   A(0,0) * (A(1,1)*A(2,2) - A(1,2)*A(2,1))
+                        - A(0,1) * (A(1,0)*A(2,2) - A(1,2)*A(2,0))
+                        + A(0,2) * (A(1,0)*A(2,1) - A(1,1)*A(2,0));
+
+    if (std::abs(det) >= 1.0e-15)
```
pbauman: Can you please make the tolerance `std::numeric_limits<libMesh::Real>::epsilon*10` or something similar? That way if libMesh::Real changes under-the-hood this also changes.

tradowsk: I didn't even know that was a thing...

Sure, I'll update it and squash it in today

tradowsk:
The change has been squashed in, and everything is still working on the large HEX27 mesh, so this should be good to go

PR #572, opened 2019-02-23
Add Multigrid examples

bboutkov:
This PR adds a couple of examples that demonstrate some multigrid capabilities for a forced Poisson problem, and for a Stokes problem using GMG on the fieldsplit velocity block.

Note that the functionality in here relies on the merging of [ libMesh PR #2047 ](https://github.com/libMesh/libmesh/pull/2047#).

Eventually, these will be used in some attempts to capture the expected number of iterations for some simple problems in an attempt to avoid regressions.


roystgnr:
Huh.  How did this pass CI when it wasn't being run against a post-2047 libMesh?

bboutkov:
Perhaps I'm mistaken but I don't think (all?) the examples are run by make check in GRINS, actually I have a branch laying around with some fixes including an example that doesn't run at all; I'll clean it up for a PR shortly. 

This PR was mainly an example that a user could run as proof of concept for the GMG implementation, I was going to add actual regression testing in the next PR as I hadn't developed the regression functionality yet, but wanted libmesh #2047 to have at least _something_.  So I'm guessing CI just installed the files to their locations and was satisfied enough.

bboutkov:
Ok, I added updated the regression_testing_app to capture the number of iterations which will now automatically be verified. I chose to depart a little bit from the current philosophy of how the app worked by instead running the simulation which generated the solution data internally to the app as the iteration counts are only accessible internal to the solver, and only once libMesh #2057 is merged.

Please let me know if there's anything that can be improved or needs fixing!

pbauman:
I've merged #577 so @bboutkov rebase on master please and then these tests can run again? I'll merge if everything passes.

bboutkov:
Hmm, this passes with PETSc 3.8.4 and that libMesh commit locally for me. Upon closer inspection it looks like CI grabbed the second to last commit? No clue why, but I think Ill just try again.

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/3170/) on 8aac619 : invalidated by @pbauman

@bboutkov pointed out that we probably didn&#39;t have git submodule update --init in our build scripts. Invalidating this and hopefully it should run now.

pbauman:
Job [GRINS-devel](https://femputer.eng.buffalo.edu:8000/job/3170/) on 8aac619 : invalidated by @pbauman

Might help if I rebuild the correct libMesh module

Issue #575, opened 2019-04-03
unit_driver test doesn't work in parallel


pbauman:
There are several tests that fail when I try to run the `unit_driver` (the CppUnit driver) test in parallel. This is annoying as it would be nice to unit test my use of non-blocking consensus to parallel-sync my overlapping fluid-solid map (among other things).

PR #576, opened 2019-04-05
Refactor OverlappingFluidSolidMap

pbauman:
Updated the API to be more user friendly and avoid gross data types. Also now fixed the lack of parallel syncing which was ultimately an issue downstream for getting a `GhostingFunctor` to work correctly. This uses `libMesh::Parallel::push_parallel_vector_data`, which now uses the NBX (non-blocking consensus) algorithm. We'll see if the older libMesh used the same API before this was updated on the libMesh side. If not, then we'll bump the minimum libMesh version (which we're going to do for GMG anyway). ATM, we don't have any unit testing of the parallel_sync (see #575).

### DIFF
--- a/src/physics/include/grins/overlapping_fluid_solid_map.h
+++ b/src/physics/include/grins/overlapping_fluid_solid_map.h
@@ -54,6 +54,10 @@ namespace GRINS
 
     virtual ~OverlappingFluidSolidMap(){};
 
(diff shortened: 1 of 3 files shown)


pbauman:
Yup. Looks like we need to bump to at least libMesh/libmesh@2a96106fd82ecf48ff44ad9bbd398dfadc4775c2 for the parallel_sync.

pbauman:
Rebased on master. Will merge if these tests pass.

PR #577, opened 2019-04-12
Update libMesh Minimum Version, Copyright Year, Other Source Cleanup

pbauman:
#576 and #572 need a newer version of libMesh. I'm pushing the minimum libMesh version to following the merge of libMesh/libmesh#2088 since we'll need that for the variational IBM stuff coming down the pipe anyway and for which #576 is intended to support.

I've also updated the copyright year 2019 in the head. Finally, there was some lingering `GRINS::` used inside GRINS namespace that I cleaned up in another branch so I just cherry-picked on this one.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,7 +3,7 @@
 //
 // GRINS - General Reacting Incompressible Navier-Stokes
 //
-// Copyright (C) 2014-2017 Paul T. Bauman, Roy H. Stogner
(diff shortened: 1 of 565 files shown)


pbauman:
OK, I've already updated the "libMesh min GRINS version" on femputer, so I'll let these run to sanity check and then I'll merge.

PR #581, opened 2019-04-17
Use node_ptr vs get_node

pbauman:
get_node was deprecated and is now being removed in libMesh/libmesh#2104
so it's time to remove the deprecated function calls.


jwpeterson:
Thanks! Do you mind invalidating the GRINS-dbg target libMesh/libmesh#2104 as well?

pbauman:
I did, it'll run in a little bit.

PR #588, opened 2019-05-22
Slightly Refactor Incompressible Fluids

pbauman:
There was actually a bug in Stokes::mass_residual. Getting rid of a bit of code duplication. I would like to template on dimension similar to what happened in solids, but that is a larger project with less priority.


roystgnr:
Wait, seriously??  What was the bug?  I can't make it out from the diff.

pbauman:
> Wait, seriously?? What was the bug? I can't make it out from the diff.

Diff is huge so it's hard to make out (aside from the many other valid reasons you have).

[Here](https://github.com/grinsfem/grins/pull/588/commits/9255f930f7698862f72f6ee8042074cabeb28510#diff-16a82c6eb58a26b2f441ecb4eebde5c7L355)

needs to be `elem_solution_rate_derivative()`.  I think it's just leftover from once-upon-a-time when `TimeSolvers` were refactored and nobody did Stokes, so we never noticed. (We noticed when monkeying with FSI for test cases).

PR #589, opened 2019-05-29
Add Thermally Coupled GMG Examples

bboutkov:
This PR expands the GMG multigrid coverage by adding a new thermally coupled flow example which uses the fixed up solver configuration settings that were part of the changes in libMesh/libmesh#2151; as such this PR requires the bumping of minimum libMesh version to that corresponding to the libMesh PR.

The content here provides a new example that is run twice with the idea of demonstrating broad categories of solver configuration capabilities. The first run uses recursive Schur complement decomposition to solve the coupled problem, while the second run uses a multiplicative fieldsplit variant with GMG to precondition the velocity subblock, and GMG on the thermal convective/diffusive subblock.

Please let me know if there's anything to be improved upon!


pbauman:
Looks good, thanks. Next we need a script generator for the command line solver options, ha!

bboutkov:
> Next we need a script generator for the command line solver options, ha!

Seriously! These things are already taking up pages and it would get even worse if we start trying to do GMG on the upper/lower components of the Schur complement inversions!

PR #590, opened 2019-05-30
Miscellaneous Fixes and Improvements

bboutkov:
Here's a collection of various minor issues I found which were all too trivial to fix on their own so I just created this branch to fix everything all at once. 

Most of these changes are cosmetic. Although the coupled Stokes-NS fixes do include new solver settings as previously the example didn't converge with default options. Also, the simple prop example was modified as it was also previously non-functioning. I wasnt too familiar with the propeller example so I'm not sure what the expected outcome was supposed to be of the simulation, but at least now its running and converging to something.


pbauman:
Thanks for these!

PR #592, opened 2019-11-06
Include missing parallel headers

bboutkov:
I haven't been quite up to date with recent libMesh changes but upon using master and trying to build GRINS I ran into a few compilation and linker errors.

I noticed some refactoring in libMesh#2219 which seems like the likely culprit, although I didn't verify this explicitly. Just thought Id submit this PR to save anyone else the potential trouble I ran into.


roystgnr:
Damn, yeah - I went through MOOSE then to add in any missing headers there, but forgot to do the same with GRINS and this must have been after the CI server was down.  Thanks!

pbauman:
Thanks! Hopefully we can revive CIVET in the not too distant future.

PR #593, opened 2020-04-09
Libmesh compatibility updates

roystgnr:
This should be backwards compatible for quite a ways, and fixes compilation for me with newer libMesh versions.


roystgnr:
So *technically* there wasn't a single CI failure!

I've already done "make check" manually (the whole point of this build was to get more test coverage of recent libMesh changes) and everything built and passed.

pbauman:
Thank you very much. Ill make it a priority to get CIVET running on my Linode so we can poke at this again.

PR #599, opened 2021-01-07
Can't std::swap two different kinds of smart ptr

### LINKED ISSUE #598: Can not be compiled with libMesh 1.7.0-pre
I tried to compile it alongside libMesh 1.7.0-pre and the problem started from this line:

../../src/physics/src/overlapping_fluid_solid_map.C: In member function 'void GRINS::OverlappingFluidSolidMap::swap_old_solution(GRINS::MultiphysicsSystem&)':
../../src/physics/src/overlapping_fluid_solid_map.C:287:93: error: no matching function for call to 'swap(std::shared_ptr<libMesh::NumericVector<double> >&, std::unique_ptr<libMesh::NumericVector<double> >&)'
     std::swap( unsteady_solver->old_local_nonlinear_solution, system.current_local_solution );
 
while I have tested with libMesh 1.6.0 and there was no problem.

roystgnr:
And in libMesh after 1.6, those two vectors are two different kinds of
smart ptr.

We'll rely on NumericVector::swap() here instead.  That means we need
the full header, too, not just a forward declaration.

This fixes compilation with the latest libMesh git head for me; hopefully it fixes #598 too.

### DIFF
--- a/src/physics/src/overlapping_fluid_solid_map.C
+++ b/src/physics/src/overlapping_fluid_solid_map.C
@@ -31,6 +31,7 @@
 
 // libMesh
 #include "libmesh/elem.h"
+#include "libmesh/numeric_vector.h"
 #include "libmesh/parallel_sync.h"
 #include "libmesh/unsteady_solver.h"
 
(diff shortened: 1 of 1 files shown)


pbauman:
LGTM, thanks @roystgnr for taking care of this.

PR #603, opened 2021-07-18
Remove deprecated input options from MeshBuilder

pbauman:
And cleaned up a few other code smells.

### DIFF
--- a/src/solver/src/mesh_builder.C
+++ b/src/solver/src/mesh_builder.C
@@ -84,18 +84,18 @@ namespace GRINS
       }
 
     // Create UnstructuredMesh object (defaults to dimension 1).
-    libMesh::UnstructuredMesh* mesh;
+    std::shared_ptr<libMesh::UnstructuredMesh> mesh;
 
     // Were we specifically asked to use a ParallelMesh or SerialMesh?
     {
       std::string mesh_class = input("Mesh/class", "default");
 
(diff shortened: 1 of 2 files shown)


src/solver/src/mesh_builder.C
```diff
@@ -98,20 +81,18 @@ namespace GRINS
       }
 
     // Create UnstructuredMesh object (defaults to dimension 1).
-    libMesh::UnstructuredMesh* mesh;
+    std::shared_ptr<libMesh::UnstructuredMesh> mesh;
```
roystgnr, line 84: Why shared_ptr and not unique_ptr?  We'd expect to have one clear "owner" of the mesh throughout its lifetime, no?

pbauman: Yeah, it's an atavism. It should be a unique_ptr. I think when this code was written originally, there was no UniquePtr wrapper in libMesh, but there was a SharedPtr wrapper that later got converted to std::shared_ptr. I think. It was a while ago.

pbauman: I didn't change it now because that will touch other pieces of code. Was trying to keep the PR self-contained.

roystgnr (approved):
Still curious about that old design decision, but the new changes here look like a nice cleanup.

Issue #600, opened 2021-01-07
Vector conversion error


ehsanakrami:
https://github.com/grinsfem/grins/blob/3c4e2ae2c7f053a41a5e99993049d797046e87e0/src/utilities/src/distance_function.C#L640

It seems that this line needs an amendment to:
```std::vector<long unsigned int> dof_ind;```

pbauman:
Hi @ehsanakrami I'm not sure I follow. What amendment do you think is needed? I might anticipate you mean that the vector needs to be sized, but the `dof_indices` call on the next line will `clear()` and `insert` into that vector so no explicit sizing/initialization is needed.

ehsanakrami:
Hi @pbauman. I have updated my comment. Sorry if it was not clear at first.

roystgnr:
That ought to be dof_id_type, right?  Not hard-coded to unsigned int *or* unsigned long, since libMesh could be configured to use either?

ehsanakrami:
Hi, Unfortunately, my libMesh couldn't and I amend it by myself to compile.

roystgnr:
Fixed by #604

PR #604, opened 2021-07-22
Fix builds with recent --disable-deprecated and 64-bit libMesh

roystgnr:
I'm still seeing test failures but at least the core code all compiles with those options again.


pbauman:
THANKS! The `unique_ptr` update was on my list for this weekend. dafa0a1 also fixes #600.

pbauman:
LGTM

PR #605, opened 2021-07-25
Adding C++11 isms

pbauman:
Started going down the rabbit hole of adding `override` last weekend and ended up also going down the rabbit hold of using `default` in many places, using `delete` on many of the default constructors we don't want to allow to be used. This led to the "using inherited constructor idiom", removing `return;` code smells and some whitespace. I did this pretty incrementally, but I couldn't squash all the commits together cleanly and I didn't think it was worth the time to try and go back and do that, so there's a few default keyword and delete keyword commits that aren't squashed.


examples/inflating_sheet/pressure_continuation_solver.h
```diff
@@ -36,6 +36,8 @@ namespace GRINS
 
     PressureContinuationSolver( const GetPot& input );
 
+    PressureContinuationSolver() = delete;
```
roystgnr, line 39: Is this necessary?  The default constructor shouldn't be implicitly defined in classes with a user-defined constructor, right?

src/physics/src/boussinesq_buoyancy.C
```diff
@@ -37,13 +37,6 @@
 
 namespace GRINS
 {
-
-  BoussinesqBuoyancy::BoussinesqBuoyancy( const std::string& physics_name, const GetPot& input )
-    : BoussinesqBuoyancyBase(physics_name,input)
```
roystgnr, line 42: This is great.  I'd forgotten how annoying writing some of these shims was.  And I don't think all the annoyance was from GRINS; I need to go through the libMesh sources to look for opportunities to use inheriting constructors there too.  I didn't even think to do that when I was skimming our sources looking for good spots to use delegating constructors.

examples/inflating_sheet/pressure_continuation_solver.h
pbauman: I did not know this. (I sanity checked on a small repro.) Thanks for pointing it out. Now to slog through and clean all of those up. (The downside of developing everything in the open is my ignorance is on full display and, therefore, the resulting shame is not private.) Really probably ought to do a pass better enforcing rule-of-five, but that's for another PR and another day.

pbauman:
Deleted all the unnecessary `= delete` default constructors (which turned out to be all of them, lol). Squashed in another `= default` on a destructor and saw a few more inheriting constructors that could be used so dropped those in da6f285; certainly not exhaustive, was just opportunistic.

pbauman:
Just to clarify, I think this is good to go if you're good with it @roystgnr

PR #607, opened 2021-08-08
Fixes for libMesh built with `--disable-deprecated`

pbauman:
The first 10 commits are relatively simple "thing was missing", like a `get_nothing()` or a `get_phi()` from the `Physics` `init_context` calls. The real work here was in the QoI classes that is after those 10 commits.

The issue was that QoIs only know about the variables they operate on, but the `System` has other variables, potentially. So my solution was to add a `register_active_vars` API so that the QoI class could indicate which of its element or side variables are active and then the `CompositeQoI` can then see what variables are inactive and then `get_nothing()` for those element and side variables. This solution doesn't feel great, but I couldn't come up with a better solution.

I also made `QoIBase::init_context` pure virtual while I was at it. There's also some other `CompositeQoI` changes, like using `std::unique_ptr`.

This gets `make check` 100%  with `--disable-deprecated`, but still a few things I want to address/think about:
1. There are some `Physics`, particularly solid mechanics, where we `get_phi()` or `get_dphi()` for the side FE, but that really depends on the boundary conditions. We need to come up with a way to have the context be aware of the boundary conditions to see what, if anything, needs to be built for the side FEs, for example. Can open an issue to track that discussion/ideas.
2. We should make `Physics::init_context()` pure virtual as well, but it didn't compile out-of-the-box so we'll address that elsewhere since this PR gets `make check` 100% passing.
3. This is without any of the chemistry libraries. Cantera's newer versions have changed things so that needs to be updated and I haven't tried Antioch yet.
4. I'd really love to update adding the QoIs, etc., on the libMesh side to avoid the `clone()` API. I really don't like these `clone()` APIs and I got caught with a copy-constructor bug I had to fix as part of this PR that took some time. It'd be nice to just hand libMesh a `unique_ptr` for these objects and not have to worry about it.
5. Since I'm breaking the API for `QoIBase` here anyway (with `init_context` becoming pure virtual and the new `register_active_vars` function), I'd love to update `QoIBase` to not have qoi_indices in the API anymore. They are unnecessary since we've moved to `CompositeQoI` interfacing with libMesh and we could update the API accordingly internally. That would be my preference anyway.


src/physics/src/scalar_ode.C
```diff
@@ -56,35 +56,44 @@ namespace GRINS
     // we've clearly got a System to grab hold of with all it's
     // variables initialized.
 
-    libMesh::ParsedFEMFunction<libMesh::Number> *tdf
-      (new libMesh::ParsedFEMFunction<libMesh::Number> (*system, ""));
-    this->time_deriv_function.reset(tdf);
+    {
+      std::unique_ptr<libMesh::ParsedFEMFunction<libMesh::Number>> tdf =
+        libmesh_make_unique<libMesh::ParsedFEMFunction<libMesh::Number>>(*system, "");
```
roystgnr, line 61: I'm still not a fan of using `auto` *everywhere*, but it's nice for idioms like this.

roystgnr (approved):
No obvious mistakes.  If you weren't running the tests manually I'd fear I was missing something subtle, though.  Have you tried `make check` with threads?

src/physics/src/scalar_ode.C
pbauman: Oh that's a good point. I also tend to err on the side of not using `auto` but agreed this could be a good place for it. It didn't trip my annoyance like iterators do (which is where I use `auto` typically).

pbauman:
That's a good point. I haven't tried this with threads. I'll do that and see. I'd be interested in your thoughts on both the design here and the questions/issues I posed, but I understand if you don't have time.

roystgnr:
1. We never should need to `init_context()` until shortly before we need to use a context, by which time we should have the boundary conditions ready too, because otherwise we'd be dumbfounded when we got to a side loop.  So can't we use the same information in `init_context()` that we'd use shortly afterward in the kernel?

2. This is suspicious.  We must have some Physics kernels somewhere that aren't implementing `init_context()`, which either means they aren't doing physics or means they're doing physics and it's only dumb luck that they're never tested outside a multi-physics app where another kernel already requested everything they need.  (or maybe smart luck?  perhaps stabilized-component kernels are the only ones not doing `init_context()`, because their corresponding unstabilized kernels already did it?

4. Avoiding clone() will never and should never happen.  We need to clone() internally to make multiple contexts for multiple threads.  A single unique_ptr won't cut it, unless we make very method const, and real const not C++ "but I can really make stuff mutable wink wink right?"; and if you have a functor simple enough for that to work then it should be trivial to copy construct.

5. I haven't looked at this in a while and don't have time to now, but it sounds reasonable.  That's really an atavism, and won't interfere with creating more flexible QoI classes, right?  When I did composable kernel objects in the past it was useful to be able to use the same kernel with different indices, or copies of the same kernel with distinct indices, rather than e.g. hard-coding variable names, but there's no reason stuff like that belongs in the base class APIs.

pbauman:
> We never should need to init_context() until shortly before we need to use a context, by which time we should have the boundary conditions ready too, because otherwise we'd be dumbfounded when we got to a side loop. So can't we use the same information in init_context() that we'd use shortly afterward in the kernel?

This was my thinking, more or less. This might require attaching some info to the `AssemblyContext`. Can carry on discussion in a separate issue.

> This is suspicious. We must have some Physics kernels somewhere that aren't implementing init_context(), which either means they aren't doing physics or means they're doing physics and it's only dumb luck that they're never tested outside a multi-physics app where another kernel already requested everything they need. (or maybe smart luck? perhaps stabilized-component kernels are the only ones not doing init_context(), because their corresponding unstabilized kernels already did it?

I didn't dig deeply, but definitely seemed to be secondary things like Boussinesq where the "primary" would've already init'ed everything necessary.

> Avoiding clone() will never and should never happen. We need to clone() internally to make multiple contexts for multiple threads. A single unique_ptr won't cut it, unless we make very method const, and real const not C++ "but I can really make stuff mutable wink wink right?"; and if you have a functor simple enough for that to work then it should be trivial to copy construct.

Oh right, threading. *sigh*

> I haven't looked at this in a while and don't have time to now, but it sounds reasonable. That's really an atavism, and won't interfere with creating more flexible QoI classes, right? When I did composable kernel objects in the past it was useful to be able to use the same kernel with different indices, or copies of the same kernel with distinct indices, rather than e.g. hard-coding variable names, but there's no reason stuff like that belongs in the base class APIs.

Cool, I can follow up on this in a separate PR.

src/physics/src/scalar_ode.C
pbauman: Add `auto` here in d173d7ac

pbauman:
> Have you tried make check with threads?

So this was a fun rabbit hole. Turns out we've not really set this up properly at all (because we always just ran with MPI). There are a few issues here that I don't think will get resolved in this PR:

1. `LibmeshInit` creates its own `GetPot` object internally for the command line parsing foo, including `--n_threads`.
2. This is fine except now the GRINS `GetPot` object does not see this so we scream and die about an unused variable.
3. Possible solution is to just parse it in GRINS and pass it to the `LibmeshInit` constructor, but maybe there are other hooks I don't know about @roystgnr?
4. I've not systematically gone through the test shell scripts to add a `GRINS_TEST_OPTIONS` or something to allow us to slam all the tests with `--n_threads=27 --warn-only-unused-var`. :( I need to come up with a better framework to lower the overhead of all those `.sh` files. It's really untenable ATM.

Ultimately, I tried one test with a variety of thread numbers. Most passed. The ones that didn't were tolerance errors for the L2 error on an exact solution. We set tolerance to `1.0e-10` and we're getting something like `1.0e-8` in those failing cases. I'm not sure if we want that to hold up this PR or not nor am I sure it's indicative of real failure or not.

roystgnr:
> Possible solution is to just parse it in GRINS and pass it to the LibmeshInit constructor, but maybe there are other hooks I don't know about @roystgnr?

For just this one argument, it would be simple enough for GRINS to query it itself and just ignore the value ... but yikes, this is a much bigger problem, isn't it?  We also have command-line overrides for `gdb`, `node-major-dofs`, 8 arguments affecting our console I/O, 7 or 8 to enable/disable various features, 2 affecting signal/exception handling, 3 for node partitioning options, 3 or 4 for solver options...  And many of these are options that don't get queried until we're already halfway into setting our first solve, so at what point can we actually safely turn on UFO detection?

> We set tolerance to 1.0e-10 and we're getting something like 1.0e-8 in those failing cases.

That's almost certainly reasonable.  We might want to enable quad precision and crank down some solver tolerances and double-check at some point, though.

pbauman:
>  but yikes, this is a much bigger problem, isn't it?

Oiy. Yeah. I'll open a separate issue to track this.

> That's almost certainly reasonable. We might want to enable quad precision and crank down some solver tolerances and double-check at some point, though.

Good idea. 

I'll go ahead and merge this and we track these two things separately.

PR #613, opened 2022-04-29
Add missing m4 macros

roystgnr:
I didn't even notice that these hadn't been added to #611

We really need to get GRINS back under CI.

### DIFF
--- a/m4/common/ax_cxx_compile_stdcxx.m4
+++ b/m4/common/ax_cxx_compile_stdcxx.m4
@@ -1,0 +1,1005 @@
+# ===========================================================================
+#  https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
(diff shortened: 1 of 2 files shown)


pbauman:
They should've been pulled in from autoconf-archive?

roystgnr:
> They should've been pulled in from autoconf-archive?

If you mean the copies I'm adding here: yes.  If you mean "automatically when we run bootstrap" somehow, not as far as I can tell.  The next time configure runs it just complains that bash can't find a command AX_CXX_COMPILE_STDCXX_17.

pbauman:
> If you mean "automatically when we run bootstrap" somehow, not as far as I can tell

I'd guess you need to `apt install autoconf-archive` (I do not know why it's not installed by default with autoconf). But if you want to merge this, I don't have any issue otherwise.

roystgnr:
> I'd guess you need to apt install autoconf-archive

You guessed correctly.  Thanks!

We ought to list that among dependencies here, but so long as it's working there's no need to fork their files.

PR #614, opened 2022-04-29
Use std::make_unique, not libmesh_make_unique

roystgnr:
The latter is deprecated now.


pbauman:
Awesome, thanks for cleaning that up!

PR #615, opened 2022-04-29
Use vertex_average (what "centroid" used to give)

roystgnr:
libMesh Elem::centroid was deprecated, since its name didn't match its
behavior.


pbauman:
Thanks for cleaning that up!

PR #620, opened 2022-05-10
Libmesh updates

roystgnr:
This gets us building again with libMesh master.  Breaks us with old libMesh, though; I wish I had time to think of a smoother upgrade path on the deprecation of direct System::qoi access and the old init_qoi() virtual.


pbauman:
Updated README with new required minimum version of libMesh. Thanks!

PR #621, opened 2022-11-25
Use GhostingFunctor::map_type rather than unordered_map

pbauman:
This captures the changes made upstream in libMesh in the API that were made for performance reasons in libMesh/libmesh#3431.


roystgnr (approved):
Thanks!

Issue #622, opened 2022-11-25
Spalart Allmaras Test Failing


pbauman:
<pre>
Stack frames: 11
0: libMesh::print_trace(std::ostream&)
1: libMesh::MacroFunctions::report_error(char const*, int, char const*, char const*)
2: libMesh::DofMap::variable_group(unsigned int) const
3: libMesh::System::get_info[abi:cxx11]() const
4: libMesh::EquationSystems::get_info[abi:cxx11]() const
5: libMesh::EquationSystems::print_info(std::ostream&) const
6: GRINS::Simulation::print_sim_info()
7: GRINS::Simulation::run()
8: /home/pbauman/work/grins/build/master/devel/test/.libs/test_turbulent_channel() [0x412db7]
9: __libc_start_main
10: /home/pbauman/work/grins/build/master/devel/test/.libs/test_turbulent_channel() [0x413d9e]
[0] ./include/libmesh/dof_map.h, line 1943, compiled Nov 24 2022 at 11:13:00
</pre>

`git bisect` points at libMesh/libmesh#3407, specifically libMesh/libmesh@b663ec5375 

I haven't dug any deeper, at least wanted to capture the git bisect results.

pbauman:
Looks like `Variables` are getting dropped.

Input:
<pre>
[Variables]
   [./Velocity]
      names = 'u v'
      fe_family = 'LAGRANGE'
      order = 'FIRST'
   [../Pressure]
      names = 'p'
      fe_family = 'LAGRANGE'
      order = 'FIRST'
   [../TurbulentViscosity]
      names = 'nu'
      fe_family = 'LAGRANGE'
      order = 'FIRST'
[]
</pre>

Program echo:
<pre>
 EquationSystems
  n_systems()=1
   System #0, "flow"
    Type "Implicit"
    Variables={ "u" "nutil" }
    Finite Element Types="LAGRANGE", "JACOBI_20_00"
    Infinite Element Mapping="CARTESIAN"
    Approximation Orders="FIRST", "THIRD"
    n_dofs()=662
    n_local_dofs()=662
    n_constrained_dofs()=0
    n_local_constrained_dofs()=0
    n_vectors()=2
    n_matrices()=1
    DofMap Sparsity
      Average  On-Processor Bandwidth <= 9.96375
      Average Off-Processor Bandwidth <= 0
      Maximum  On-Processor Bandwidth <= 10
      Maximum Off-Processor Bandwidth <= 0
    DofMap Constraints
[...]

PR #623, opened 2024-04-06
Update header, exact solution tols

pbauman:
Some header cleanup upstream in libmesh we needed to account for in 7e3d4d51b25d43444bcb427eeb8fba21af8d80a2.

We also update some exact solution tolerances due to update libMesh/libmesh@e8cec186f9206db4318ae50315e695d3413c0d9c which updated the default quadrature order in the exact solution error estimation. (`git bisect run` is soooooo satisfying when it works)

This now gets us passing all tests except the turbulence one, which is known in #622.


roystgnr (approved):
Thanks!

PR #624, opened 2026-02-17
libMesh compatibility updates

roystgnr:
These are sufficient to get me passing all tests with `--disable-deprecated` builds of libMesh `devel`.


roystgnr:
And it's still working with libMesh today, and I need to add a new feature or two for testing; let's merge this.

